Issue #194, opened 2014-04-01
Windows installer crash my environment PATH


mojocorp:
Under windows 7, the installer crash the windows PATH environment variable....

renebohne:
Can't we just skip the Winsows Installer? The ZIP version runs fine, right?

mgmax:
No, the windows installer is easier to use, just like you want a nice 
package on Linux. We use it on our Windows machines.

t-oster:
hmm.... don't like the registry... but may be the cleanest windows-like solution. What about a pull-request ?

mgmax:
@t-oster the issue can now be tagged low priority, because since #249  the installer detects the situation and then does not touch PATH anymore.
I don't consider the case of ultra-long PATH important because we support until 1024 and if you go only a little further, you will hit the windows limit of 4096 (at least in the Win7 GUI).

mgmax:
I would suggest closing this issue, as the inkscape plugin runs fine without PATH and VisiCut doesn't destroy the PATH variable anymore if it is too long.

t-oster:
if somebody objects, please reopen

Issue #174, opened 2013-08-29
Scaling 1.25 off


dirkx:
Am trying to find out why visicut is often 1.25 for SVG files generated on the mac (with for example OmniGraffle - which generally display right in other tools).

Simples of examples is below - I'd expected this to be 10x10mm; it shows up as 0.8 or 0.81 depending on the phase of the moon.

Dw.

`<?xml version="1.0"?>
<!DOCTYPE svg PUBLIC "-//W3C//DTD SVG 1.1//EN" "http://www.w3.org/Graphics/SVG/1.1/DTD/svg11.dtd">
<svg xmlns="http://www.w3.org/2000/svg" xmlns:xl="http://www.w3.org/1999/xlink" version="1.1">
  <defs/>
  <g stroke="none" stroke-opacity="1" stroke-dasharray="none" fill="none" fill-opacity="1">
    <g>
      <rect x="5mm" y="5mm" width="10mm" height="10mm" stroke="black" stroke-linecap="round" stroke-linejoin="round" stroke-width="1"/>
    </g>
  </g>
</svg>`

renebohne:
I created an SVG file with the rectangle specified by dirkx:
http://hci.rwth-aachen.de/~bohne/simpleRectDPITest.svg

If I open it in Illustrator CS5, the coordinates and dimensions are ok.
In Inkscape, te stroke width leads to (x,y)=(9.5mm,186,5mm) and dimensions (101mm,101mm).

In VisiCut 1.7-98 on Mac OS X, it is imported as (7.857,7.857) with dimensions (80.262 mm,80.262 mm)

renebohne:
If I build from sources (develop branch) and start the build, this version of visicut imports the rect as follows:
(x,y): (14.189 mm, 14.189 mm)
(w,h): (143.579 mm,143.579 mm)

Does the build server use a different version of some libraries (e.g. svgsalamander) than the ones included on github?

mgmax:
another problem is if one edits an illustrator-svg with inkscape. In this case VIsiCut should warn and ask "Did you last edit this SVG with Inkscape or Illustrator?", instead of assuming that it is an Illustrator-SVG and making the scaling wrong.

peteruithoven:
Let's only show warnings when where quite sure we can't figure this out using meta data in the file?

t-oster:
[...]

Issue #171, opened 2013-07-26
Severe UI malfunction under Mac OS X 10.7.5


DavidMenting:
Since installing a Java update on OS X 10.7.5 (now Java version 1.6.0_51) Visicut has a severely malfunctioning user interface. It starts empty and only redraws when you resize the window:

![screen shot 2013-07-26 at 16 24 04](https://f.cloud.github.com/assets/869645/862995/ff66d7ce-f5ff-11e2-8329-71d53953efb4.png)

Log entries look like this:

26-07-13 16:27:51,072 JavaApplicationStub: CGContextGetCTM: invalid context 0x0
26-07-13 16:27:51,072 JavaApplicationStub: CGContextSetBaseCTM: invalid context 0x0
26-07-13 16:27:51,072 [0x0-0x38f38f].com.t_oster.visicut.gui.VisicutApp: Jul 26 16:27:51 Davids-MacBook-Air.local JavaApplicationStub[6761] <Error>: CGContextGetCTM: invalid context 0x0
26-07-13 16:27:51,073 [0x0-0x38f38f].com.t_oster.visicut.gui.VisicutApp: Jul 26 16:27:51 Davids-MacBook-Air.local JavaApplicationStub[6761] <Error>: CGContextSetBaseCTM: invalid context 0x0
26-07-13 16:27:51,073 JavaApplicationStub: CGContextGetCTM: invalid context 0x0
26-07-13 16:27:51,073 JavaApplicationStub: CGContextSetBaseCTM: invalid context 0x0
26-07-13 16:27:51,073 [0x0-0x38f38f].com.t_oster.visicut.gui.VisicutApp: Jul 26 16:27:51 Davids-MacBook-Air.local JavaApplicationStub[6761] <Error>: CGContextGetCTM: invalid context 0x0
26-07-13 16:27:51,073 [0x0-0x38f38f].com.t_oster.visicut.gui.VisicutApp: Jul 26 16:27:51 Davids-MacBook-Air.local JavaApplicationStub[6761] <Error>: CGContextSetBaseCTM: invalid context 0x0

The program is unusable. Dialog boxes don't pop up and it is impossible to click buttons.

I might try a downgrade of the installed Java version to see if Visicut starts working normally again.

mgmax:
http://informagen.com/JarBundler/StubFile.html suggests symlinking the java stub.

t-oster:
Looks like a solution. Who can test if this works? I would like the OSX build to work on most OSX versions possible.

Issue #165, opened 2013-07-05
Windows installer does not detect JRE on Windows 7/8


t-oster:
Appearantly there are issues when installing on Windows 7 or 8, that the installer does not detect the installed JRE (Java Runtime Environment). It then tries to install one, which also fails and the aborts the installation in general.

Since I do not have access to any Windows 7 or 8 installations and not much time, I would appreciate it very much, if someone could take a look into that issue.

The related files are in "distribute/windows" (https://github.com/t-oster/VisiCut/tree/develop/distribute/windows) and the installer is written using the Nullsoft Scriptable Installer System (NSIS) (http://nsis.sourceforge.net/Main_Page). 

Take a look at the distribute.sh script to see, how the executable is created:

```
echo "Creating Windows installer"
  [ -d wintmp ] && rm -rf wintmp
  mkdir wintmp
  cp -r windows/* wintmp/
  [ -d wintmp/stream ] || mkdir wintmp/stream
  cp -r visicut/* wintmp/stream/
  cat windows/installer.nsi|sed s#VISICUTVERSION#"$VERSION"#g > wintmp/installer.nsi
  cp ../tools/inkscape_extension/* wintmp/
  cat ../tools/inkscape_extension/visicut_export.py|sed 's#"visicut"#"visicut.exe"#g' > wintmp/visicut_export.py
  pushd wintmp
  makensis installer.nsi > /dev/null || exit 1
  popd
  mv wintmp/setup.exe VisiCut-$VERSION-Windows-Installer.exe || exit 1
  rm -rf wintmp
```

PR #228, opened 2015-01-15
Release Version 1.8

t-oster:
This Pull-Request shall be the place for discussion for the 1.8 release. I think it is time to do this release, because there are many changes since 1.7.

Please comment which Bugs/Features you think have to be resolved before this release should be done.

I will update this description with the things to do:
## TODO:
- [ ] select and fix important bugs:
  - [ ] #174 Scaling 1.25 off
  - [x] Bugs specific to Windows
    - [x] #165 Windows installer does not detect JRE on Windows 7/8
    - [x] #194 Windows installer crash my path
  - [ ] Bugs specific to OSX
    - [ ] #171 Severe UI malfunction und Mac OS X 10.7.5
- [ ] update changelog


renebohne:
Can we make sure that SVG import scaling is free of bugs before we release 1.8?
I think that issue #174 is a very important thing, because people in our fab lab not only create SVGs with Illustrator or Inkscape, but with any software. We have many problems with scaling and need to be very careful before we actually start cutting.

renebohne:
These features that have a deadline of 4.2.2015:
thingiverse integration #221 
automatic arrangement of shapes #222 
facebook integration #223 
SVG Exporter #225 
TSP Optimizer #226 
Usability Improvements #227 

Since they are developed by students, we need to have some time for testing... but if the release can wait until 1.3.15, I would love to see these features in the major version 1.8 :)

mgmax:
My opinion is that we should skip the upcoming features, because they will inevitably have bugs that do not appear immediately.
The release should be as stable as possible - if users are interested in new features, they can just use the develop version.

renebohne:
ok... so what are we waiting for? What has to be done for 1.8?
To me, only issue #174 is a serious thing that is nothing I would expect in a stable release. 
What are the other things?
#171 : I can't reproduce this bug anymore. Mac OS X 10.9.5. Maybe someone with 10.10 can test if there are problems with it? Otherwise, I think that this bug got fixed...

I can't say anything about the two windows bugs right now, but I will install windows this weekend and give it a try.

mgmax:
@t-oster please update the description. #165 is fixed, the core problem of #194 is also fixed.

README.md
```diff
@@ -9,8 +9,8 @@ www.visicut.org
 
 Download
 --------
-* Latest stable release: http://137.226.142.25:5080/wwwshare/VisiCutNightly/master
```
brains12345: hoping etching problem solved

mgmax:
As it's been ages since the last release, I would suggest switching to a "rolling release" scheme:
In a certain time interval, the development branch is frozen and declared as release candidate. A pull request then allows for discussion and feedback by beta testers. If after maybe three weeks there is positive feedback and no regressions were discovered, we now have a new release.

After some time, this can be repeated for the next release. 

As a contribution, I would offer beta-testing these versions at FAU FabLab, as we are currently using random snapshots of the development branch.

mgmax:
It's been ages since the last release, and all recent changes are bugfixes. I bet the current development version is better than the previous release in (almost?) all aspects and that the two 'release blocker' bugs #171 and #174 were already present before.

I think it's time to release; or just automatically set the version number to (year-2000)/10, which would mean releasing 1.8 in 2018 :-)

t-oster:
I think the separation of "develop" aka "unstable" and master aka release does not make sense anymore. I am merging develop to master and bump the version to 1.8. If we have really experimental features, they can be compiled in a separate branch as testing version.

PR #230, opened 2015-01-15
file import: positions and dimensions are now displayed

renebohne:
file import: positions and dimensions are now displayed in the INFO message after a modification of the file contents.

Not sure if you need the .xml files...

In previous versions, VisiCut only said that something changed after a file was imported. It was hard to find out, what scaling factor was applied. Now, the user can see how big his drawing was before and what changes VisiCut made. No more frustration after importing a file :)


t-oster:
cleaned up and merged in fe121370205a2fb135f654c2f6a89f01b9ee5a73. XML files are not important.

Issue #227, opened 2015-01-14
Usability Improvements: keyboard shortcuts


renebohne:
A group of students will add support for the most important keyboard shortcuts:
command + c, 
command + v, 
delete

t-oster:
command+D for duplicate is already there and delete should also work, if I remember correctly.
copy/paste is not implemented, therefore no shortcuts exist.

renebohne:
true, but copy and paste will be added with this issue :)
Delete does not work on my Mac.

ksughosh:
The command + D option doesn't work. But creating duplicates by right clicking works. However there the new duplicate overlaps on the previous. Hence this occlusion prevents the user from getting the feedback if the part was duplicated. 

It would be better if the duplicated part had some offset from the original so that we could easily distinguish whether the duplicate has been generated. This has been addressed here:

https://github.com/t-oster/VisiCut/pull/236

Also the duplicated part is now on top and is the selectedPart.

Issue #226, opened 2015-01-14
TSP VectorOptimizer


renebohne:
One group of students implemented a TSP VectorOptimizer.
At the moment, it is based on scpsolver.

We all know that TSP is a hard problem and that the optimisation will need resources and a lot of computing time. But the result will improve cutting time.
Therefore, we need to make sure, that the result is saved into an SVG or PLF file, so that the calculation can happen at home and the results can be opened in the fab lab (using sorting "FILE").

I will attach some images later...

Problems:
- Only 25% better than NN approximation
- Depending on path segmentation
- Exponential running time

renebohne:
![tsp1](https://cloud.githubusercontent.com/assets/671105/5736193/2f3d804a-9bd3-11e4-82b0-c4fd740a2d8c.jpg)
![tsp2](https://cloud.githubusercontent.com/assets/671105/5736194/3369f2d4-9bd3-11e4-8acb-e4a5652d0f03.jpg)
![tsp3](https://cloud.githubusercontent.com/assets/671105/5736195/374b2ac6-9bd3-11e4-92b0-881f79aa4b75.jpg)

mgmax:
sounds nice, but in practice it is also very important that inside holes are cut first.

Most SVG files have connected paths, where the current implementation of INNER_FIRST sorting works. If a path is split, we have problem #203 that the sorting is theoretically correct inside-out, but a huge waste of time.

It would be interesting to see how large the potential for improvement actually is - I doubt we can reach more than 10% in real-world files and that the "path-optimal solution" is optimal to what the user wants.

t-oster:
I cannot see the TSP as optimizer in VisiCut, after (locally) merging #241 . Is this feature fully implemented?

feklee:
*What's the status of the optimizer?*

I'm running 1.7-245-g89a10cd.

Remarks and questions:

  * From my experience with a Trotec Speedy 300, the software for which supports sorting vectors, speedup can easily be 30%. Today I've witnessed a pattern being cut on an Epilog Zing, and the laser was jumping from one end to the other all the time.

  * Cutting the inner holes first should only be necessary for specific designs with small pieces that relocate when lose. TSP would be optional, anyhow, and also one could order by layer or color. 

  * Ordering by centers of gravity of curves would be a big improvement in many cases. No need to solve the traveling salesman problem.

mgmax:
I have a completely opposite opinion regarding "inner holes first": I 
[...]

Issue #225, opened 2015-01-14
SVG Exporter


renebohne:
Some students are writing a liblasercut SVGExporter "device driver". 
It will allow to export the current elements (after vector optimization) into an SVG file.
A file save dialog will pop up if the user hits "execute"

Some tricky issues might be:
* layers for multiple input files and mappings

mgmax:
You could reuse code from the dummy driver, which has a SVG export and 
playback functionality:
https://github.com/t-oster/LibLaserCut/blob/develop/src/com/t_oster/liblasercut/drivers/Dummy.java

Issue #206, opened 2014-10-25
minor usability improvements


mgmax:
While watching users I noticed some minor issues.

(edit: some of these are now fixed, see the checkboxes)

 regarding "Open with Visicut" from the inkscape plugin (probably the same for Illustrator):
- [x] automatically fetch new image from camera when a new file is opened
- [ ] when the preview is zoomed in, sometimes the new file is not visible because it is outside the zoom area - in this case, zooming out automatically would be nice
- [x] the VisiCut window should signal the window manager that it wants to pop into foreground ("steal focus") - at least under Linux+KDE it doesn't work or is somehow prevented by the system.

camera image: (a really great feature of VisiCut!)
- [x] hide / gray out old image when a new one could not be fetched
- [x] somehow mark the image as "old" after pressing Execute? (probably not a good idea because it can be annoying in some situations)
- [x] close previous "cannot load image" error boxes on clicking "fetch new image"

regarding drag/resize:
- [ ] if the object to cut has a very small size, one cannot drag/resize it without zooming in far enough, because all handles are overlapping
- [x] users sometimes mistake the X and Y offset for width and height, because they are shown very near the bounding box 
- [x] black font for height/width/X/Y is not readable on all backgrounds - draw a drop shadow or 50%-transparent white area behind it?

All these are no real problems, I hope we @fau-fablab can do a "day of code" sometime and fix these.
Thanks a lot for VisiCut!

PR #236, opened 2015-02-01
VisiCut Optimization and Usability Improvements

ksughosh:
Here the following issues have been addressed.

- https://github.com/t-oster/VisiCut/issues/225
- https://github.com/t-oster/VisiCut/issues/226
- https://github.com/t-oster/VisiCut/issues/227
- https://github.com/t-oster/VisiCut/issues/206

The SVGExport driver has been integrated, and the LibInfo has been updated accordingly. However it doesn't show up on list of devices. Maybe this would serve as an initial implementation.

The TSP vector optimizer for VisiCut has been integrated. Their respective dependencies are also updated accordingly.

The keyboard shortcuts for copy, paste, delete have been implemented. The copy paste now actually copies the exact image and pastes it at some offset so that there is visual feedback of the copy operation taking place.

The second usability improvement was when the panel is zoomed in, loading a new file will zoom out the interface back to 100%. Hence there is proper feedback that the file has been loaded.


t-oster:
Work at #241

Issue #104, opened 2013-03-12
Visualize execution order


mgmax:
It is not immediately clear from the mapping dialog that the order in the mapping also represents the order in which the profiles are executed.
This:
![screen5](https://f.cloud.github.com/assets/1520891/250880/f4613c0c-8b56-11e2-80b7-9c9e90caea45.png)

cuts first and engraves second, which is usually not what the user wants.

Suggestions:
- add a number "1,2,3" before each line to make it clearer
- put the order into the lasersettings tab (so that engrave is always before cut no matter how it is sorted in the mapping, except when you change it there)

mgmax:
Can it be that we are horribly over-complicating the whole issue? Taking a look at 3D printer software like Cura, we should just make it simple and do the right thing ™. If someone wants to implement the really advanced selections nicely, he can do so, but obviously there is not enough time for that. As long as there are so many other bugs, we should rather choose the simple solution that makes 95% of the users happy.

![Cura Screenshot image](https://cloud.githubusercontent.com/assets/1520891/9701457/4f8c1f4a-542a-11e5-8943-2014987ef55f.jpg)

I would be happy with just merging everything and forcing the order to the default one that you set up somewhere in the settings.

PR #237, opened 2015-02-03
Automatic Object Arrangement

C4rmel:
We are still working on this, but you can have a look at it.


peteruithoven:
I would really love to see a screenshot

C4rmel:
Well a screenshot of this is not really spectacular since you only see arranged shapes :)

One interesting screen can be when the objects that you want to arrange do not fit all into the laserbed. We the try to maximise the used area and mark the pieces that can be removed to achieve this.

objectarrangerscreen

peteruithoven:
Nice! 
So this arranges multiple files right? Not shapes in files right? 
If it's per file it might make more sense to put the Arrange button next to the file editing ui (File selector, add file button, remove file button). Because now it's grouped with view related buttons. I'll admit I'm not sure where to place it when it's not per file. 
I also assume the button only shows up when relevant? To keep the ui as minimal as possible for beginners.

C4rmel:
thanks for the feedback!

Yes it is per file. In the beginning we tried to do it for shapes in files. But this has several problems since most SVGs usually are not grouped correctly so the arrangement process might separate parts that usually belong together. Also as far as I got this visicut strips down the SVG files into parts and does not maintain the group relation. We also got the feedback that the general case where this option would be useful is when you dump several files into VisiCut and it would be too tedious to arrange them by hand. So we made the decision to only arrange files.

Regarding the button: I need to admit that we did not put that much thought process into the actual position but more in the function behind it and what happens when you press it. We highly focused on the slider interface that you see on the right side. (Does not seem like a lot of work I know) but the whole interaction was way more complex before and so we did some user test and stripped down the whole interaction to this one slider that appears after you pressed the arrange button. We did test this with novice and experienced users. I will consider your feedback regarding the button position because you are right that it belongs more to the file interaction. We wrote a report for the Plugin and will also give a  talk where we talk about our design decisions and show the iterations of the interface and previous prototypes. As far as I know the talk gets recorded. I can put linkshere for the report and the talk when they get published.

peteruithoven:
Really great to hear that you did user research. I do like the simplicity of the slider interface you guys created.
I'd love to see the talk.

mgmax:
Did you try cutting and engraving multiple files at once on an epliog laser?
There will probably be a problem (although it is not your code to blame):
When I last tried this engraves the first item, cuts the first, engraves the second, cuts the second, and so on. Because of a driver limitation the job can only contain exactly one engrave part and then multiple cut parts. (TODO check this with the windows driver and colour mapping in the "advanced" tab)
Related: #104

renebohne:
No status update for 5 days... are there problems with the things that Thomas requested?

C4rmel:
Hey I am pretty interested in finishing this thing but at the moment I am realy busy. I am writing my bachelor thesis and I need to learn for exams. I can continue this project after the 17.03.

t-oster:
Hi... any update on this?

C4rmel:
I have to admit I forgot about this. I got in touch with my partner for this. We both are busy on September (exams and thesis month) but we plan to finish this on the beginning of October. We also got some more feedback from the final presentation at our chair.

t-oster:
Please reopen if you find the time to continue.

PR #241, opened 2015-02-10
Ksughosh correct (clean version. fixes #236,#240)

t-oster:
Here a rebased version. We should do some testing before merging, because I did not yet look at the code, just at the commit-structure).


t-oster:
So.... anyone interested in this feature? If not, I am going to remove it, because I cannot see an improvement, which would justify adding a big library and bloating the code

mgmax:
In my opinion, the code would need some improvements before it could be merged... For example, variable names like "test7" and "juhu" are not very helpful.

I think that the TSP is more academic and not for practical use, because it takes too long to solve for larger files. We should still have a look on some of the related commits in this PR, like improving copy-paste. 

Am 16. Dezember 2015 09:55:52 MEZ, schrieb Thomas Oster <notifications@github.com>:

t-oster:
Nobody seems interested in the TSP so I close this. If anyone has the time to integrate the other features, please create a new PR.

PR #275, opened 2015-06-04
GUI changes

FroChr123:
Additional changes in GUI, see seperate commits for more details.


mgmax:
Moving it into a menu is a good idea :+1:
I would suggest renaming it to "Share", so that users immediately know what's inside.

At least under Linux the GUI is now slightly too wide for 1024x768.
On 1280x1024 - most of our lab's PCs - it works.


compare:
before (1024 width)
before

after (1024 width)
after_1024

after (1280 width)
after_1280

FroChr123:
Thanks for the feedback @mgmax, I could not test all different systems.

For the current selection of plugins renaming it to "Share" would make sense, yes. But there are some plans to add some "non-sharing" plugins currently, so this was a bit future-oriented. :)

I am mostly confused about the resized progress bar in your picture with 1024 pixels, it was intended to be set to a fixed width and for my testing scenario it is working correctly. I will try to have a closer look at this and why this happens, this was not intended.

FroChr123:
System dependant bugs should (hopefully) be fixed with:
https://github.com/FroChr123/VisiCut/commit/3432d7e6686608e20f32972b492fb2fc9340b36c

Feedback is appreciated. :)

mgmax:
The problem with "Plugins" is that the user does not know what to find there. These features aren't even plugins because they are part of the main installation and you can't download and install them from some kind of "app store" like a browser plugin.

Features not related to sharing should be put into a separate menu.

peteruithoven:
I have to agree with mgmax. When it isn't installed by the user as a plugin, a plugins menu doesn't really make sense.

FroChr123:
I will simply rename it to Share and put the future changes in this menu aswell, although these new changes do not match with Share completely, but they still have somthing to do with Sharing. Another menu for a single menu item would just blow up the GUI again and I currently cant think of a better name.

PR #276, opened 2015-06-05
GUI, camera, projector

FroChr123:
For details, please see seperate commit comments.


t-oster:
hi, are those features production ready? or is this a long time pull-request?

FroChr123:
The stuff above should already be production ready, but as these modifications are part of my thesis, there will be a bunch more of them. Currently working on QR code detection and after that on a webservice which connects all these different parts.

FroChr123:
Just found one thing, which is a bit production-ready-relevant (not that important but can be annoying because it causes warnings, but that whole part can be stopped by the user).

Fixed in next commit, also included the minor optional code changes mentioned above.

mgmax:
The camera "on/off" button looks a little weird because it does not fit the way other buttons work and also doesn't match the known look for "on/off" widgets. As far as I know my lab's users, they don't like learning to use non-standard widgets and need something extremely intuitive to learn that the old camera button now works different.

Labeling it with "On" is confusing, either we should use no text at all, or "Camera on"/"Camera off". As it is no alert message of global interest, it should not use a highlighting color, because no other button does so.

It is also not immediately clear if the button represents the state ("the camera is on, click here to change") or the action ("switch on the camera by pressing this").

FroChr123:
I am open to better solutions, but as already mentioned above, the default UI elements for such purposes (Checkbox, Toggle Button) are in my opinion even worse.

In fact the button has to change, with a "single capture" button there are no support possibilities for video-like previews, which are now possible. I know that users are a bit unflexible to changes, but on the other hand it is not a good idea to stick with a low performance "single photo capture" button forever and disallow any progress due to it, if there is now support for automatic video-like previews.

I chose the Icon + On / Off solution because it does not need much horizontal space, which is the reason why Camera on / Camera off is not a good solution I think. The color can be easily removed, I thought it would be a good thing to highlight that stuff, as it is important for the usage of the program.

Another idea would be to design some new icons to represent that the camera / projector are now on or off.

In general it is currently not feasible for me to spend hours of thinking about the design of a button since I have to meet very tight deadlines and there are still several hundreds / thousands lines of code to be done.

FroChr123:
I still feel that a CheckBox would be the correct solution for this as it is a default UI element for showing "state on / off" of something. But I have some doubts about using them, mainly:

1) Not very noticeable, breaks completely with current button-ish design
2) Looks weird in combination with a symbol (camera, projector), maybe fallback to text?

What are your opinions about it?

peteruithoven:
Just to clarify (I'm afraid I'm struggling with the German): 
- The view menu has 3 new items: 
  - Activate camera (enables the auto camera capture feature, instead of the manual one) (removes the actuate photo in favor of an on / off toggle button) 
  - Ignore camera (I don't understand this one?)
  - Activate projector (enables projecting the preview in the laser bed using a projector) (Adds a projector toggle button)

Have you noticed in your own experience that you often enable / disable the camera or projector feature? 
In my experience at fablabs usually users don't configure that much, they just download the settings. Is there a reason for users to enable / disable the camera / projector? 
What I'm basically getting at, I'm not sure we need enable/disable buttons above the preview. (This would solve the whole ui element struggle) 

A more difficult question. Isn't auto capturing vs manual capturing a lasercutter specific setting? That depends on the hardware and software setup per lasercutter? Are we sure this is something we can put in the view menu?

FroChr123:
Thanks for your answer @peteruithoven. :)

Let me try to explain some things first:
- Activate camera: Does exactly what you described. Once it is deactivated, the image is simply frozen and the last taken photo is shown in the preview.
- Ignore camera: This is basically the new representation of the previous menu item "Display photo" / "Foto anzeigen". I guess this feature was not used very often before that change aswell. It has the functionality to "clear" the preview and just show the simple one colored background of the chosen material, so in contrast to deactivating the camera, it actually "removes" the camera image in the preview.
- Activate projector: Does what you described.

You are making a good point in the question if we need these elements in the UI at all. The answer to that point is again a bit complicated, but I will try my best to explain the whole situation:

All these changes are related to the idea of using QR codes in the workflow of FabLabs, so you can basically identify projects / cutting schemes with the help of QR codes. You can show them to a webcam to simply load them or use a VisiCam setup to place / rotate the QR codes on the material, which you want to cut and a projector is supposed to display these cutting lines directly on the material. This is the whole concept / idea behind all these changes.

With that in mind, lets get back to your question, if we need the UI elements at all:
- Projector: I think for the projector it would only make sense to disable the projector preview if the preview interferes with the scanning of the QR codes. Currently I can not tell if that situation is likely to happen or not, but I assume that the projector enabling stuff is not "really" needed as an UI element.
- Camera: The detection of QR codes is obviously directly connected to the camera status. As soon as you have placed the QR codes correctly on the material, you need to be able to "freeze" the camera image, such that you can remove the QR codes from the laser cutter before you start the actual laser cutting job and VisiCut keeps the detected positions. In this context the enabling / disabling of the camera is rather important, so an UI element is definitely needed for this.

Basically the projector UI element is not that important. For the camera I can currently only think of one solution to bypass a "permanent" UI element for the enabling / disabling of the camera: As soon as a QR code is detected an "information message" is displayed. This message already contains a simple button to disable the camera, so deactivating is no problem. But we also need to be able to enable the camera easily. Maybe it is a good idea to automatically turn on the camera again after a laser cutting job was started?

To answer your last question: What this change currently basically does is, that it just frequently asks automatically for a new image. If your software / hardware setup supports refresh times in the range of milliseconds, you will get a new image with that timing as well. If your setup only supports a refresh rate of some seconds, you will get a new image after that period of seconds as long as the camera is activated. If the camera is deactivated, it simply stops asking for new images. I think with a manual capturing button we can only support manual and slow capturing, but with that "camera activated / deactivated" setting as it is currently both setups are supported.

Additionally to that, there are settings for the refresh rates of the camera and projector, per lasercutter. With these settings you can artificially slow down the refresh rates (e.g. your system supports 70ms refresh rates, but you decide that you only need 500ms).

I have experimented a bit with the checkboxes. They do not look as bad as expected, but they are not very satisfying aswell.

visicut-checkbox

What do you think? Is it a good idea to remove the projector stuff from the main UI (still available in "View") and to use a message box to display the "disable camera" button if it is needed (= after QR codes are detected) and automatically re-activate the camera after the laser cutter was started? Of course the camera should still be reachable in the "View" menu, e.g. if a user removed the message box (which contained the easily accessible button to deactivate the camera).

This solution would currently look similar to this:
visicut-deactivate-camera

peteruithoven:
I don't fully understand the QR code concept. Instead of loading designs in VisiCut you show a QR code? And instead of selecting a cutting scheme (is that the mapping + profiles?) you show another or the same QR code? Instead of placing / rotating the design you place a design using it's QR codes? Do you have your project explained anywhere online? 

If I understand the workflow, a user doesn't really want to stop the camera, the thing he wants to do is store the positions? 
Maybe a button that says "store location" would make more sense? This could also be placed in the information message you mention. Maybe if you place it in the message, you don't need it in the top bar? Maybe just storing the location and not stopping the camera makes more sense? 

In the preview you display the stored positions in a slightly different color, or show the currently detect locations slightly more transparent? 
I wouldn't remove the stored location after starting a job because you might want to cut on the same location again. (When doing "production", where you cut a whole sheet of material and then want to cut another whole sheet for example)

I like the auto capturing functionality. 
I would put the following items in the menu:
- Enable camera. Disabling this would hide the camera image. I believe checkboxes that disable something are confusing (because you enable a disable or you disable an enable).
- Enable auto capturing.  When this is disabled a capture button appears in the top bar (like it is currently implemented). I personally don't think a auto capture checkbox is needed in the top bar. If you think it is needed I think adding the checkbox in the topbar, like you've tried looks fine. 
- Enable projector. Enables / activates the projector functionality. Again are you sure we need a checkbox in the top bar?

FroChr123:
The QR codes actually just store a URL to a file, which can be loaded by VisiCut (for example SVG or PLF). As soon as such a QR code is detected by VisiCam, it downloads the file and loads its center position to the center position of the QR code and applies the rotation as it is currently shown by the QR code.

Unfortunately, the files have to be loaded without any mapping currently. I guess you have already noticed that changing the mapping of a file e.g. from "cut" to "engrave" takes a very long time and the responsible code for that is currently way to inefficient to use this in combination with the QR codes. Since I have to meet the deadline, I do not have the time to re-write this whole code for higher efficiency, which could be used in combination with the QR codes.

There are plans to add a more detailed project description to the website of the HCI (http://hci.rwth-aachen.de/) in the future.

You are right with the point that actually deactivating the camera is just used to store the locations of the QR codes. Renaming might be a good idea.

But I think there is one issue with the functionality of the button, if we introduce a state "locations stored / locations not stored" and still let the camera turned on. Turning off the camera is actually used to ensure that the situation of the QR codes in the image will not change anymore. If the camera is still turned on after the locations are stored, this might easily end up in inconsistent results, let me try to explain this a bit more detailed.

Technically it currently works like this:
- Store current objects, which were loaded by QR codes
- Detect all QR codes in the image
- Load files, load objects to correct positions / rotations
- Delete old stored objects

So it basically throws away all loaded QR codes and loads them completely new again frequently. I think better solutions would require some kind of QR code "tracking" mechanism, which would be computationally and regarding the available amount of time way beyond everything which is currently possible.

If we introduce a state, which says that QR code locations are saved and still have the camera turned on, we run into two issues:
- If "Locations stored" does not block the loading of new QR codes from the image, we might easily end up in multiple unwanted copies of the same QR code, because it gets loaded multiple times (= no QR code tracking available).
- If "Locations stored" blocks the loading of new QR codes, it is easily confusing for the user, because there are no possibilities to change this state and if the user shows a new QR code to the system (and sees that image in VisiCut with the new QR code), it would simply be ignored, which is confusing I think: Users show a new QR code, it is displayed in the preview, but it will never be recognized as new QR code in that situation.

All in all: You are right that deactivating the camera is only used to store the location. But I think it can not be actually implemented in that way without having some kind of QR code tracking mechanism, which is highly complicated. So I think we unfortunately have to stick with deactivating the camera for this purpose (it is also some kind of feedback for the user that the image is frozen), but what do you think, should we still rename the button in that situation?

The top bar currently does not need the camera / projector stuff anymore, because this is either not important enough (projector) or it is shown dynamically when needed (camera). Both items are still reachable with the "View" menu.

QR code loaded stuff is loaded without any additional stuff currently, it is loaded as a normal file, without mappings or any other modifications, mainly because of performance reasons (the file loading itsself already takes a larger amount of time than expected, but it is still in the range of feeling interactive). Threaded file loading or cached file loading are both not possible with the current file loading system.

Cut positions multiple times: You are right, first idea, which comes into my mind:
Ask user with a Yes / No dialog, if he wants to reload the positions / keep positions after cutting?

Enable camera: The ignore camera stuff was already a bit unintuitive at the beginning, when it was called "Display photo / Foto anzeigen". Combining it is a good idea I think. Will only be available in "View" menu or for freezing the QR code image.

Auto capturing: Good idea, but I think putting it in the "Preferences" tab is even better, because it is a rather permanent setting, which would never be changed I think.

Enable projector: Will be available in the "View" menu only.

peteruithoven:
> Technically it currently works like this:
> - Store current objects, which were loaded by QR codes
> - Detect all QR codes in the image
> - Load files, load objects to correct positions / rotations
> - Delete old stored objects
This happens when the Deactivate camera button is pressed? Or continuesly when the camera is enabled?

FroChr123:
This happens continously when the camera is enabled. It is the quickest solution, which does not need a fully working QR code tracking mechanism.

peteruithoven:
With tracking you mean it can follow the movement of QR codes that where found? Isn't there a library for that? I've created a library years ago for OpenFrameworks. You can't be the first to try this in Java? 

Can't you create 2 QR code "storages"? 
- One where you store the last found QR codes in the image.
- One where that last found QR code is copied to when a user presses something like *Store locations*.

FroChr123:
I bet there are libraries for it. But I am not sure that these libraries are fast enough in processing the stuff and this is a critical aspect of the whole stuff, processing times should be as small as possible. Lets put it that way: In the whole context of this project I also had to use some external libraries of course and some of them promised stuff, which they were not really able to fulfill. In combination with the limited amount of time and deadlines, I think it is the better tactical decision to get the stuff running with a basic support and worry about the optimizations later - there are always things to be optimized, but unfortunately there is not enough time to deal with all the possible optimizations.

t-oster:
Hi,looks like you are making many changes to visicut for your project/thesis which is an interesting idea by the way. However, since visicut is used by many people we have to ask if those changes make sense for the main branch or are more a scientific proof of concept/research prototype,which should be kept in a different fork until the Features make sense for most users.

Dont get me wrong,i like it when people are using the code but i want to prevent unnecessary bloating of the code.

FroChr123:
Hi, yeah I know that this is another thing of modifying VisiCut.

I think the biggest change for a user of VisiCut (who is not interested in using the QR codes and has not enabled it in the options) would be the automatic camera image capturing instead of clicking the capture photo button. Anything else should behave exactly as before.

Codewise I try to keep the changes in the original files as small as possible, as you can imagine some (mostly smaller) things in the main code files needed to be changed. With these changes I try to keep the stuff in seperate files, so that the new code is still managable and not spread around all original files.

I did not pull request the latest changes yet as well, because the new changes are currently in an unfinished state. As soon as everything is done I think I will create another pull request and then we could check again if it makes sense to merge it in the main branch or keep it in the seprate fork. :)

FroChr123:
I think that I am done with the biggest parts of my modifications, the code is here:
https://github.com/FroChr123/VisiCut

I was also able to implement the stuff a bit better, which mainly means three things:

1) "Preview" QR codes are now clearly visibly shown in VisiCut with green and thick lines

2) Store locations: As already mentioned above, tracking QR codes would be a bit complicated, but I was able to achieve a similar effect with a less accurate method, which should work perfectly for our situation: As soon as positions are stored, the position of the QR code is stored as well and in a small radius around this position, the same QR code will be ignored as long as there is any PlfPart of this QR code left in the preview. This avoids the issue, that a QR code would be loaded multiple times on the same position after storing the location.

3) Mappings: Mappings are still removed in the preview, but as soon as positions are stored, the mapping is loaded (as mentioned above, caused by performance constraints for the preview).

What do you think, is this currently fine for a pull request?

PR #287, opened 2015-07-21
Added IllustratorScript for Windows (Fixes #286)

### LINKED ISSUE #286: Illustrator extension for Windows
I think it would be really useful to have an Illustrator extension that also works on Windows. Currently, because it's written in AppleScript it only works on Macs. So I thought VBScript would be easy to make it work on Windows.

I don't really know a lot about programming so I worked on the script itself using some Adobe documention but not on the installer through VisiCut. Like how to install the Inkscape extension you go to Extras>Install Inkscape-Extension. Anyways, here is the code for the script:

     'Exports the current document as SVG to dest
     'dest contains the full path and file name to save to

     destFile = "C:\temp\VCtemp.svg"
     exportFileAsSVG (destFile)

     Sub exportFileAsSVG (dest)
          Set appRef = CreateObject("Illustrator.Application")
          Set svgExportOptions = CreateObject("Illustrator.ExportOptionsSVG")
 
          If appRef.Documents.Count > 0 Then
               svgExportOptions.EmbedRasterImages = True
               svgExportOptions.FontSubsetting = 7 'aiAllGlyphs
               Set docRef = appRef.ActiveDocument
               Call docRef.Export (dest, 3, svgExportOptions) ' 3 = aiSVG
          End If
     End Sub

     'open document in VisiCut

     strFileName = "C:\temp\VCtemp.svg"
     Set oShell = CreateObject("WScript.Shell")

     oShell.Run "visicut.exe "  & strFileName

If you save this code as a .vbs and place it in your Illustrator's \presets\scripts folder it should work, but I'd like to be able to install it like you do with the Inkscape extension. After that you could just select the extension from the scripts menu of Illustrator.

On NetBeans, I've been looking into the Helper.java file where I find the instructions for Mac Installation. So what's my next step?

t-oster:
This will add a windows-compatible version of the Adobe Illustrator Extension to the menu.

TODO:
 * [ ] Test if the installation works
 * [ ] Add (x64 and x86) Paths for all important Versions of Illustrator or find a method to determine the path of an installed version
 * [ ] Test the script on all important versions of Illustrator

### DIFF
--- a/distribute/distribute.sh
+++ b/distribute/distribute.sh
@@ -37,6 +37,7 @@ cp ../tools/inkscape_extension/*.py visicut/inkscape_extension/
 cp ../tools/inkscape_extension/*.inx visicut/inkscape_extension/
 mkdir -p visicut/illustrator_script
 cp ../tools/illustrator_script/*.scpt visicut/illustrator_script/
+cp ../tools/illustrator_script/*.vbs visicut/illustrator_script/
 echo "Compressing content..."
 [ -f VisiCut-$VERSION.zip ] && rm VisiCut-$VERSION.zip
 zip -r VisiCut-$VERSION.zip visicut/  > /dev/null || exit 1
--- a/src/com/t_oster/visicut/misc/Helper.java
+++ b/src/com/t_oster/visicut/misc/Helper.java
@@ -310,7 +310,8 @@ public class Helper
       new File("/Applications/Adobe Illustrator CS6/Presets"),      
       new File("/Applications/Adobe Illustrator CS4/Presets.localized"),
       new File("/Applications/Adobe Illustrator CS5/Presets.localized"),
-      new File("/Applications/Adobe Illustrator CS6/Presets.localized")
+      new File("/Applications/Adobe Illustrator CS6/Presets.localized"),
+      new File("C:/Program Files/Adobe/Adobe Illustrator CC 2015/Presets.localized"),
     })
     {
       if (dir.exists() && dir.isDirectory())
@@ -658,12 +659,12 @@ public class Helper
 
   private static File getIllustratorScript()
   {
-    return new File(new File(getVisiCutFolder(), "illustrator_script"), "OpenWithVisiCut.scpt");
+    return new File(new File(getVisiCutFolder(), "illustrator_script"), isWindows() ? "OpenWithVisiCut.vbs" : "OpenWithVisiCut.scpt");
   }
 
   public static boolean isIllustratorScriptInstallable()
   {
-    return isMacOS() && getIllustratorScript().exists();
+    return (isMacOS() || isWindows()) && getIllustratorScript().exists();
   }
 
   public static String allowedChars = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789.-";
--- a/tools/illustrator_script/OpenWithVisiCut.vbs
+++ b/tools/illustrator_script/OpenWithVisiCut.vbs
@@ -1,0 +1,24 @@
+'Exports the current document as SVG to dest
+'dest contains the full path and file name to save to
+
+destFile = "C:\temp\VCtemp.svg"
+exportFileAsSVG (destFile)
+
+Sub exportFileAsSVG (dest)
+Set appRef = CreateObject("Illustrator.Application")
+Set svgExportOptions = CreateObject("Illustrator.ExportOptionsSVG")
+
+If appRef.Documents.Count > 0 Then
+     svgExportOptions.EmbedRasterImages = True
+     svgExportOptions.FontSubsetting = 7 'aiAllGlyphs
+     Set docRef = appRef.ActiveDocument
+     Call docRef.Export (dest, 3, svgExportOptions) ' 3 = aiSVG
+End If
+End Sub
+
+'open document in VisiCut
+
+strFileName = "C:\temp\VCtemp.svg"
+Set oShell = CreateObject("WScript.Shell")
+
+oShell.Run "visicut.exe "  & strFileName


t-oster:
example builds http://download.visicut.org/feature-illustrator-win

diegohuyke:
I've tested this with Illustrator CC 2015, the latest version, and it works well.

I tested the feature-illustrator-win visicut build but it did not work, instead I got a success notification on VisiCut but nothing happens (the .vbs file is not moved to the location).

The closest I've been to getting the installer to work is replacing the code for the Illustrator installer with code copied over from the Inkscape extension. In other words they install the same way. This is on line 667 of my code:

     private static File getIllustratorScript()
      {
         try
        {
           String path = Helper.class.getProtectionDomain().getCodeSource().getLocation().getPath();
           if (path == null)
           {
            return null;
           }
           String decodedPath = URLDecoder.decode(path, "UTF-8");
           File folder = new File(decodedPath);
           return folder.isDirectory() ? folder : folder.getParentFile();
        }
         catch (UnsupportedEncodingException ex)
        {
          Logger.getLogger(Helper.class.getName()).log(Level.SEVERE, null, ex);
        }
        return null;
      }

And then the placing of the file is done like this:

     public static void installIllustratorScript() throws IOException
      {
         File src = new File(getIllustratorScript(), "illustrator_script");
         if (!src.exists() || !src.isDirectory())
         {
            throw new FileNotFoundException("Not a directory: "+src);
         }
         File trg = null;
         if (isWindowsXP())
        {
             trg = new File(new File(FileUtils.getUserDirectory(), ".inkscape"), "extensions");
         }
        else if (isWindows())
        {
            trg = new File (new File (new File (new File(new File(new File (System.getenv("systemdrive"), "Program Files"), "Adobe"), "Adobe Illustrator CC 2015"), "Presets"), "es_ES"), "Secuencias de comandos");
        }
        else
       {
          trg = new File(new File(new File(FileUtils.getUserDirectory(), ".config"), "inkscape"), "extensions");
       }

       if (!trg.exists() && !trg.mkdirs())
       {
           throw new FileNotFoundException("Can't create directory: "+trg);
       }
       for (File f :src.listFiles())
       {
           if ("OpenWithVisiCut.vbs".equals(f.getName()))
       {
           File target = new File(trg, "OpenWithVisiCut.vbs");
        BufferedReader r = new BufferedReader(new FileReader(f));
        BufferedWriter w = new BufferedWriter(new FileWriter(target));
        String line = r.readLine();
        while (line != null)
        {
          if ("VISICUTDIR=\"\"".equals(line))
          {
            line = "VISICUTDIR=r\""+getVisiCutFolder().getAbsolutePath()+"\"";
          }
          w.write(line);
          w.newLine();
          line = r.readLine();
        }
        w.flush();
        w.close();
        r.close();
      }
      else if (f.getName().toLowerCase().endsWith("vbs"))
      {
        FileUtils.copyFileToDirectory(f, trg);
      }
    }
  }

I noticed that this would mean that the Windows XP and the  Mac installation then wouldn't work. I don't know how to fix the Windos XP, but I'm sure the Mac bit can't be that hard if instad of going to user.home (getUserDirectory) we could go to the applications folder.

My problem is that when I click the Install Illustrator-Extension in VisiCut, I get an error saying that it couldn't copy it to the destination and then it lists the correct file directory followed by (Access Denied). Can the .java file place a .vbs in a directory under /Program Files/?.The other problem I notice with this is that I don't know how to do the localized instalation. For now I just have it set up for Spanish instalations of Illustrator.

t-oster:
Well first your getIllustratorScript method returns VisiCuts basepath, so you should use getVisiCutFolder() instead and only append the "illustrator_extension"...

A normal program (without admin rights) won't be able to put files in "program data" on windows > XP i think. Please google if there are also illustrator-extension-directories in the users home. I bet there are some, so different users can have different extensions installed.

Maybe they differ for different OS, so you should use the correct path for each OS like in the inkscape-install method.

diegohuyke:
I've looked into it and it turns out Illustrator only can load scripts onto the menu if they are under that one directory in /Program Files/. It can, however, really open scripts from any directory if you go to Scripts>Other scripts. This let's you select a script from any folder which is why I'm assuming they dont care for having user specific scripts.

I was actually able to make the code above work if you open up the .jar with admin rights (this seems to solve the access denied issue). Basically I split the code into a `If (Helper.isWindows())` part and and `Else` part. The only problem is that I'm not sure still how to fix the localization issue, right now it's all configured for Spanish computers. This is for installing the script:

     public static void installIllustratorScript() throws IOException
       {
         if (Helper.isWindows())
       {
           File src = new File(getIllustratorScript(), "illustrator_script");
         if (!src.exists() || !src.isDirectory())
         {
          throw new FileNotFoundException("Not a directory: "+src);
        }
    File trg = null;
    if (isWindowsXP())
    {
        trg = new File(new File(FileUtils.getUserDirectory(), ".inkscape"), "extensions");
    }
    else if (isWindows())
    {
        trg = new File (new File (new File (new File(new File(new File (System.getenv("systemdrive"), "Program Files"), "Adobe"), "Adobe Illustrator CC 2015"), "Presets"), "es_ES"), "Secuencias de comandos");
    }
    else
    {
      trg = new File (new File (new File(new File(new File (System.getenv("/"), "Aplicaciones"), "Adobe Illustrator CC 2015"), "Presets"), "es_ES"), "Secuencias de comandos");
    }

    if (!trg.exists() && !trg.mkdirs())
    {
      throw new FileNotFoundException("Can't create directory: "+trg);
    }
    for (File f :src.listFiles())
    {
      if ("OpenWithVisiCut.vbs".equals(f.getName()))
      {
        File target = new File(trg, "OpenWithVisiCut.vbs");
        BufferedReader r = new BufferedReader(new FileReader(f));
        BufferedWriter w = new BufferedWriter(new FileWriter(target));
        String line = r.readLine();
        while (line != null)
        {
          if ("VISICUTDIR=\"\"".equals(line))
          {
            line = "VISICUTDIR=r\""+getVisiCutFolder().getAbsolutePath()+"\"";
          }
          w.write(line);
          w.newLine();
          line = r.readLine();
        }
        w.flush();
        w.close();
        r.close();
      }
      else if (f.getName().toLowerCase().endsWith("vbs"))
      {
        FileUtils.copyFileToDirectory(f, trg);
      }
    }
    }
    else
    {
      String errors = "";
    for (File dir : new File[]{
      new File("/Applications/Adobe Illustrator CS3/Presets"),
      new File("/Applications/Adobe Illustrator CS4/Presets"),
      new File("/Applications/Adobe Illustrator CS5/Presets"),
      new File("/Applications/Adobe Illustrator CS4/Presets.localized"),
      new File("/Applications/Adobe Illustrator CS5/Presets.localized"),
      new File("/Applications/Adobe Illustrator CC 2014/Presets.localized"),
      new File("/Applications/Adobe Illustrator CC 2015/Presets.localized")
    })
    {
      if (dir.exists() && dir.isDirectory())
      {
        for (File d : dir.listFiles())
        {
          if (!d.isDirectory())
          {
            continue;
          }
          if (!"Secuencias de comandos".equals(d.getName()))
          {
            d = new File(d, "Secuencias de comandos");
            if (!d.exists() || !d.isDirectory())
            {
              continue;
            }
          }
          try
          {
            FileUtils.copyFileToDirectory(getIllustratorScript(), d);
          }
          catch (IOException ex)
          {
            errors += "Can't copy to "+d.getAbsolutePath()+"\n";
          }
        }
      }
    }
    if (!"".equals(errors))
    {
      throw new IOException(errors);
    }
    }
      }

And this is for getting the extension:

     private static File getIllustratorScript()
       {
         if (Helper.isWindows())
       {
           try
         {
          String path = Helper.class.getProtectionDomain().getCodeSource().getLocation().getPath();
          if (path == null)
         {
            return null;
          }
          String decodedPath = URLDecoder.decode(path, "UTF-8");
          File folder = new File(decodedPath);
          return folder.isDirectory() ? folder : folder.getParentFile();
         }
         catch (UnsupportedEncodingException ex)
         {
          Logger.getLogger(Helper.class.getName()).log(Level.SEVERE, null, ex);
         }
         return null;
         }
         else
         {
           return new File(new File(getVisiCutFolder(), "illustrator_script"), "OpenWithVisiCut.scpt");
         }
      }

The rest is the same as the previous code. What do you think I should do now?

t-oster:
Well... is there a possiblity to add the location of a custom script to the menu? Maybe though a registry-entry or somtehing (just add the script and search for it's path in the registry).
It does not make sense to me to add code, which only runs with admin privileges.., except if you find a way that visicut asks for admin-rights only when installing the extension... but a registry entry for the menu sounds more reasonable to me

t-oster:
So is there still any work in progress for this pull request? Anyone who can search for non-admin ways to add scripts to illustrator on windows? Otherwise I will close this.

t-oster:
Well, since nobody seems to be working on this, this is closed.

PR #297, opened 2015-09-06
minor scripts work

jnweiger:


jnweiger:
The image of the Goldcut ABH 721 cutting plotter actually belongs to https://github.com/t-oster/LibLaserCut/pull/29, but its path resides outside of LibLaserCut. Hmmm.

PR #298, opened 2015-09-06
Improve camera error handling: Ignore 'marker not found', and more

mgmax:
- ignore 'marker not found' errors, just disable the image (only works with newest patches to VisiCam)
- ramp up the retry delay time after errors (first error: quick retry, later: wait 10 seconds more)
- handle closing the error message from the code, not via timeout. If fetching the image took very long and then failed, the error message kept disappearing and reappearing, causing annoying UI resizing.

The last point doesn't work perfecly yet, because when I tested it sometimes the error message was still open when the problem was gone. I don't know why because the code looks correct to me.


mgmax:
update: now I found and fixed the problem (pull request is already updated via amend and force push).

PR #301, opened 2015-09-08
A small test cut object, inspired by Silhouette America.

jnweiger:
It is so practical, I use it almost daily :-)

### DIFF
--- a/src/com/t_oster/visicut/gui/PreviewPanelKeyboardMouseHandler.java
+++ b/src/com/t_oster/visicut/gui/PreviewPanelKeyboardMouseHandler.java
@@ -77,6 +77,7 @@ public class PreviewPanelKeyboardMouseHandler extends EditRectangleController im
   private JMenuItem deleteMenuItem;
   private JMenuItem flipHorizMenuItem;
   private JMenuItem flipVertMenuItem;
+  private JMenuItem rot90CwMenuItem;
   private JMenuItem openMenuItem;
   private JPopupMenu backgroundMenu;
   private JMenuItem startPointSetMenuItem;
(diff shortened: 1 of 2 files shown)


t-oster:
Why did you add an extra button for that file? We already have the File->Examples menu, which contains all example files...

jnweiger:
True. Technically it does not add any functionality that was not there before.
It adds 
* to usability, as it points out one object that should be used for testing and for reference, and 
* to usability, as it makes the object a one click to encourage test cuts.
* an item to the comparison checklist. Almost all commercial tools have a 'test cut button'.

peteruithoven:
I'm all for encouraging people to make test cuts, but which file seems subjective, why not add a engraved part for example? 
Also we have `Laser Script` that can create a test file that cut's in multiple speeds and powers, which I personally feel is the best way to start off if you're unsure about your settings. See the [Test pattern wiki page](https://github.com/t-oster/VisiCut/wiki/LaserScript:-Test-pattern) or [Fablab's test patterns](https://github.com/Fablab-Amersfoort/visicut-settings/tree/master/examples) for example. But then, one problem is that the range you'd want to test depends on the material type. And this also doesn't include engrave testing. 

If we can't come to a consensus I feel the test file needs to be another setting that people can share and customize. 

I'm curious what you guys think.

mgmax:
I think Jürgen wants something like the "Test Cut button" that many 
plotters have: You just push the button and it immediately cuts a 
"circle within a square" test pattern which is perfect for all settings 
that a vinyl drag-knife type cutter has like offset and pressure.

This pattern is not particularly useful for lasercutters, you could just 
cut a rectangle or a single line as well.

I'm not sure what is the right solution, because the other test jobs for 
lasercutters also have their use and should be at a similar position in 
the menu.

PR #302, opened 2015-09-11
scroll-move-rotate improvements

jnweiger:
* keep the object in the visible area, move the area if needed.

* add a right-click-menu item for rotate 90 deg.


t-oster:
Nice. Should we also include 90° counter-clockwise? Of cause this can be done by 3 times 90°... just a suggestion.
What about internationalization?

PR #306, opened 2015-10-01
Fix for windows without terminal services

grbd:
Under the version of windows 7 I was using (Home Premium)
d=os.environ.get("SESSIONNAME") resolves to "Console" instead of None
since terminal services isn't part of this version of windows this then leaves to the failure of the command "query session"


grbd:
Just to mention I can confirm this works under Windows 7 Home Premium

mgmax:
unfortunately, "Console" is also the output if you are running terminal services and are at the physically connected screen of the machine. Could you just write a check to see if "query session" fails?

grbd:
I've amended the pull request, but it seems to be pulling in other commits now
I'll close then re-open it after merging

mgmax:
you can always force-push to 'overwrite' the branch you have submitted as a pullrequest.
But be careful as any forced operation may 'delete' data. If you know git well enough, you don't really need to care because you can always restore old stuff from the reflog.

grbd:
I did do a force push after amending the commit, but for some reason the pull request then decided to pull in other commits that had arrived after the commit I'd added at the same time

PR #309, opened 2015-10-05
Editable Job Name

jnweiger:
The Zing lasercutter only shows the first 16 characters of a job name.
The 'ink_ext' prefix or filename are sometimes insufficient to clearly
distinguish jobs when navigating the job history on the lasercutter.
Especially, when the same object is sent multiple times with different
visicut-settings (engrave, cut, ...) it would be great, to give descriptive names, what the change was.

This commit provides a way for the user to add a short memonic text information
to the job name.

This helps for later retrieval from the job history and also
when multiple users send jobs with possibly similar jobs names.

### DIFF
--- a/src/com/t_oster/visicut/gui/MainView.java
+++ b/src/com/t_oster/visicut/gui/MainView.java
@@ -1658,7 +1658,6 @@ public class MainView extends javax.swing.JFrame
     this.calculateTimeButton.setVisible(estimateSupported);
     this.timeLabel.setVisible(estimateSupported);
     this.jLabel10.setVisible(estimateSupported);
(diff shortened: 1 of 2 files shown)


t-oster:
Nice one. Maybe we should then remove the "job-prefix" setting of the zing driver?

t-oster:
If you select a lasercutter, wich cannot calculate time, the label of the job-prefix-textfield is hidden. Maybe you attached it to the wrong group?

jnweiger:
Yeah, possibly wrong group. I got a headdache puzzling the groups. Any help appreciated.

jnweiger:
Found it. There was wrong setvisible() call copy/pasted. Much better now. Please check again.

PR #316, opened 2015-11-25
Export gcode

mrzl:
Ok, this is a beginning, to get the stone rolling. I've been implementing some things according to your advice https://github.com/t-oster/LibLaserCut/issues/32

The commits for LibLaserCut are in the separate repository, for suggestions on how to handle this, let me know.

### DIFF
--- a/src/com/t_oster/visicut/VisicutModel.java
+++ b/src/com/t_oster/visicut/VisicutModel.java
@@ -56,6 +56,7 @@ import java.beans.PropertyChangeListener;
 import java.beans.PropertyChangeSupport;
 import java.beans.XMLDecoder;
 import java.beans.XMLEncoder;
+import java.io.*;
 import java.io.File;
 import java.io.FileInputStream;
 import java.io.FileNotFoundException;
@@ -63,6 +64,7 @@ import java.io.FileOutputStream;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
+import java.io.PrintStream;
 import java.net.SocketTimeoutException;
 import java.util.Arrays;
 import java.util.Enumeration;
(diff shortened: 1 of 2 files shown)


src/com/t_oster/visicut/VisicutModel.java
```diff
@@ -757,6 +757,24 @@ public void sendJob(String name, ProgressListener pl, Map<LaserProfile, List<Las
     }
   }
 
+  public void saveJob( String name, ProgressListener pl, Map<LaserProfile, List<LaserProperty>> props, List<String> warnings ) throws IllegalJobException, SocketTimeoutException, Exception{
+	  LaserCutter lasercutter = this.getSelectedLaserDevice().getLaserCutter();
```
t-oster: As the counterpart in LibLaserCut, this method should get an output stream where to put the data.

src/com/t_oster/visicut/gui/MainView.java
```diff
@@ -1789,7 +1798,96 @@ private void editMappings() throws FileNotFoundException, IOException
...
-
+	private void generateGcodeMenuItemActionPerformed( java.awt.event.ActionEvent evg ) {
+		String jobname = getJobName();
+		List<String> warnings = new LinkedList<String>();
+		final Map<LaserProfile, List<LaserProperty>> cuttingSettings = this.getPropertyMapForCurrentJob();
+		if (cuttingSettings == null)
+		{
+			return;
+		}
+
```
t-oster: Here we should ask the user, where to save the file. One thing, we did not consider yet is the file extension (not every driver outputs gcode). To make it perfect, the driver should give the file extension (e.g. via a method getOutputFileExtension())... but right now I could live with ".dat" or something like this.

mrzl:
Cheers, I adopted your thoughts, and went the easy way of throwing an exception, when a lasercutter is used that doesn't implement the saveJob() method for now.

mrzl:
By the way, a huge advantage of being able to export gcode, at least for me, using a smoothieboard, is that you can decrease the time a cutting job takes massively by uploading the generated gcode via the webinterface of the smoothieboard. It then doesn't wait for the 'ok' reply after every command. But I recon that behavior can be achieved by setting setWaitForOKafterEachLine to false, right?

t-oster:
The driver is able to upload the file via Web interface, too. Just unset Hostname and port and let the http upload url be correct. 
I think we need a wiki page for the driver

-------- Ursprüngliche Nachricht --------
Von: Marcel Schwittlick <notifications@github.com> 
Datum:29.11.2015  03:59  (GMT+01:00) 
An: t-oster/VisiCut <VisiCut@noreply.github.com> 
Cc: Thomas Oster <mail@thomas-oster.de> 
Betreff: Re: [VisiCut] Export gcode (#316) 

By the way, a huge advantage of being able to export gcode, at least for me, using a smoothieboard, is that you can decrease the time a cutting job takes massively by uploading the generated gcode via the webinterface of the smoothieboard. It then doesn't wait for the 'ok' reply after every command. But I recon that behavior can be achieved by setting setWaitForOKafterEachLine to false, right?

PR #318, opened 2015-12-02
fix DPI Adobe/Inkscape inconsistency

jnweiger:
Attention horror ahead. 

We had an drawn an object with inkscape that was shown 20% too big in visicut.
The file came originally from Adobe illustrator and uses px dimensions.

Inkscape is stupid enough to keep the Creator comment of Adobe for ever and then writes a standard 90 DPI svg with this comment still there.

My suggestion is to detect both Adobe comments and inkscape namespaces. Inkscape wins then.


t-oster:
Thanks for your contributions, always a pleasure to just read your code and merge it right away.

PR #343, opened 2016-02-18
Improved layout for lasercutter profile dialog box.

pelrun:
I've broken up the large number of fields in the lasercutter profile window to a few tabbed panels. It seems a lot less intimidating now. Also gives extra room for the settings table and the job done text field.

visicutlcpe


peteruithoven:
Nice! @pelrun the world needs more people that make things less intimidating.

PR #366, opened 2016-05-07
Replace the affine transform for camera alignment with a homography. …

hghile:
… This allows all perspective correction to be done within visicut.


t-oster:
Hi, looks very nice from the first review. I will have to set up a camera in order to test this, but I think it is nice.
However, VisiCam used to detect markers and calibrate after each image, which takes small movements of the camera into account. Now it is only calibrated once. I think it depends on your Lasercutter/Camera Setup if this works reliable.
How accurate is your calibration in practice?

mgmax:
It would be helpful for VisiCam users to have some description or suggestion to select "2" as the number of points. Otherwise there will be two homographies applied after another, which - as far as I can tell - will at best be a waste of CPU and at worst distort the already corrected image.

For reference, the previous data flow was as follows:
1. VisiCam does perspective correction and also crops the image to the area between the markers.
2. Then VisiCut does an affine transformation to take the physical scaling and also the location of the machine coordinate system origin into account.

The split between the two parts in the dataflow was mainly because VisiCut can cut a test pattern for calibration while VisiCam does not even understand that the device is a lasercutter.

hghile:
I did some test cutting following calibration and the rendered preview entirely hid the cuts of the laser.  I don't know what accuracy that is in terms of mm, but it was very good.

I tried visicam, and aside from the heavy installation burden, the accuracy with which I could place the markers at the corners of the bed and which opencv could detect them did not give good results.  And the home position of the laser carriage blocks the corners for any reasonable camera position, so in my setup it is not feasible for it to recalibrate every time.  I don't know how big an issue camera stability will be, but recalibration is only a the cost of a sheet of paper.

How important is it that visicam and visicut be split components?  Visicut does know the device is a lasercutter and it knows about a camera as input, so having it do all the processing seems OK to me.  And it's much simpler and can support standard cameras instead of requiring raspberry pi, and is faster because it isn't relying on the slower rasbperry pi for processing.  i think there shouldn't be penalty in using visicam in addition to a homography within visicut other than increased lag; the documentation for how to setup visicam seems like a good place to suggest that using 2 point affine transform is the preferred method.

pelrun:
Visicam doesn't require a Pi; you can run it on the same computer you're running Visicut if you want to.

hghile:
I guess you could run them both on one machine, but that just seems like unnecessary complexity to have to run them separately.

t-oster:
I don't think it has any disadvantages to add the homography to VisiCut because you could still use VisiCam if you want to and get the same results (if the image is corrected before by VisiCam, the Homografy in VisICut should IMHO resolve to an affine transform).
But if we already have support for the perspective correction in VisiCut, we should considder adding support for USB webcams aswell. Not every Webcam has builtin support for HTTP (which was another reason for VisiCam).

mgmax:
Does this change break old settings or are they converted to the new format?

t-oster:
As I read in the code it should convert old affine transforms automatically into corresponding homographies, but I did not test it yet

hghile:
I attempted to convert the old affine transform from settings into the new format, but as it's not a setup I was using I can't tell you with certainty that it's correct.  I can tell you that does not crash though.

mgmax:
the conversion seems to work.
Pressing cancel in the camera calibration dialog still applies the changes.

The new annotation of marker order in the camera calibration dialog by small ticks (one for the first marker, two for the second, etc.) is a little confusing at first, but still less confusing than before where the image was upside-down if you exchanged the two markers that looked exactly the same.

mgmax:
If I don't forget I will try cutting and calibrating on a real machine tomorrow.

hghile:
any chance to try it out?

t-oster:
Hi, yes I tried it and it works very nice. The marker labels could be improved of cause, but there is always room for improvement. Please rebase to current tip of develop, then I can merge it.

peteruithoven:
Could someone clarify for me how this impact usage of the webcam without VisiCam? Do I need to add markers or can I still go through the calibration proces manually?

t-oster:
The only thing this changes is, that you can use more than two markers in the camera calibration process in VisiCut, which allows to correct perspective errors. I.e. it allows for the camera image to be not perfectly perpendicular to the laser-beam.

peteruithoven:
That sounds awesome. So just to clarify, your talking about virtual / digital markers here? Not the visual / physical ones?

t-oster:
We are talking about the markers, which are physically lasered on the camera calibration page and then have to be virtually moved to match the position on the photo.

peteruithoven:
Alright, that clarified it. 
I was thinking you guys might have been talking about these kind of markers:
visicam markers

hghile:
I think i've got it synced up now

t-oster:
well, you merged develop, but rebasing creates a cleaner history. I rebased it for you and merged it. Thank you for your contribution, I think it makes the Camera feature much more useful.

PR #397, opened 2016-12-20
Do not use width="100%" height="100%" for scaling when seen in document root

jnweiger:
Fixes https://github.com/fablabnbg/VisiCut/issues/7

I am unsure, if there are other cases, where UT_PERCENT is misinterpreted. Please double check!


t-oster:
well... this does not seem to break anything in any case, so regardless of other cases, this should be merged.

jnweiger:
thanks!

Issue #347, opened 2016-03-09
Update px unit size for files created by Inkscape >0.92


pelrun:
When investigating #346 it became clear that we're going to have an issue with documents produced by Inkscape 0.92 and later, because the px unit is moving from 90dpi to 96dpi. The original value was due to a typo in the CSS2 standard that was fixed in an errata, but Inkscape used the incorrect value. It's finally been fixed in the development version.

Since Illustrator and other SVG producers use different sizes for the px unit, Visicut tries to determine the creator of an imported file and set the dpi appropriately. We'll soon need to distinguish between SVGs from Inkscape <=0.91 and Inkscape >=0.92 for this to work consistently. Also, the fallback value is 90dpi which seems like it should also be changed to 96 since non-Inkscape producers are more likely to use the latter value than the former.

jnweiger:
Start inkscape0.92 with out loading an svg file, draw some objects, and send to an extension.
The svg sent has no inkscape:version element.
Save as box.svg, then send to an extension, the svg contains 'inkscape:version="0.92.1 unknown" sodipodi:docname="box.svg">'

As an inkscape extension, we cannot know if we should assume 96 or 90 dpi, unless we find a 'sodipodi:docname' element. 
If we find sodipodi:docname, but we don't find inkscape:version, we can assume 90. 
If we find both, we can assume 96.

Looks like a bad inkscape bug to me!

PR #409, opened 2017-04-22
Fix https://github.com/t-oster/VisiCut/issues/347

jnweiger:
Inkscape 0.92 generates SVG in 96 DPI, all earlier version do 90 DPI.

### DIFF
--- a/src/com/t_oster/visicut/model/graphicelements/svgsupport/SVGImporter.java
+++ b/src/com/t_oster/visicut/model/graphicelements/svgsupport/SVGImporter.java
@@ -49,6 +49,8 @@ import java.io.InputStream;
 import java.io.InputStreamReader;
 import java.net.URI;
 import java.util.List;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 import java.util.logging.Handler;
 import java.util.logging.Level;
 import java.util.logging.LogRecord;
@@ -122,7 +124,7 @@ public class SVGImporter extends AbstractImporter
       public void close() throws SecurityException{}
 
     };
-    
+
     try
     {
       u.clear();
@@ -180,6 +182,7 @@ public class SVGImporter extends AbstractImporter
     double result = 90;
     boolean AdobeIllustratorSeen = false;
     boolean WwwInkscapeComSeen = false;
+    boolean InkscapeVersion092Seen = false;
     boolean usesFlowRoot = false;
     try
     {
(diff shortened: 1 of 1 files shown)


src/com/t_oster/visicut/model/graphicelements/svgsupport/SVGImporter.java
```diff
@@ -197,6 +200,33 @@ public double determineResolution(File f, List<String> warnings)
...
+
+	        int vers_maj;
+		try { vers_maj = Integer.parseInt(matcher.group(1)); }
+		catch (NumberFormatException e) { vers_maj = -1; }
+
+	        int vers_min;
+		try { vers_min = Integer.parseInt(matcher.group(2)); }
+		catch (NumberFormatException e) { vers_min = -1; }
+
+		if ((vers_maj >= 0) && (vers_min >= 92))
```
mgmax: I would suggest checking for  `((vers_maj == 0) && (vers_min >= 92)) || (vers_maj > 0)` so that we also handle a future version "1.xx".

src/com/t_oster/visicut/model/graphicelements/svgsupport/SVGImporter.java
```diff
@@ -237,22 +267,26 @@ public double determineResolution(File f, List<String> warnings)
...
+           warnings.add("Inkscape Version 0.92+ comment seen in SVG.");
+	 }
+       warnings.add("Switching DPI from 90 to " + result + " - Please check object size!");
     }
     return result;
   }
 
   /*
    * Tries to determine the Coordinate resolution in DPI.
    * SVG default is 90, but AI generates 72??
```
mgmax: Actually, the SVG standard says **96**, but inkscape < 0.92 wrongly used 90 dpi. http://wiki.inkscape.org/wiki/index.php/Release_notes/0.92#Important_changes

We should somehow suppress the DPI warning message for files loaded from a recent inkscape version via the inkscape extension, because users will ignore a warning that pops up every time.

jnweiger: Oh jeah, I meant to include all 1.x version of inkscape too. Thakns for catching this.

The SVG standard consortium wronly decided to change the existing standard instead of releasing a new version of the standard.

The warning messages are only intended to be there temprorarily to raise awareness of the issue. I am pretty sure there are cornercases, that we cannot catch.

mgmax: Just out of curiosity, do you have a source for the statement that the SVG standard was 90dpi earlier?

jnweiger: https://www.w3.org/Style/css2-updates/REC-CSS2-19980512-errata.html
Section 4.3.2

mgmax:
Seems to work with a few illustrator <-> inkscape crossover editing combinations that I tested: https://gist.github.com/mgmax/eb73693a04da6d1b7a9f3d9681f12473
Maybe someone wants to build a unit-test for these?

The warning messages should be improved (although this doesn't need to be a part of this pull request):
- For files with explicit (non-pixel) width/height units, there should be no warning message at all. They also work fine in VisiCut without the inkscape-0.92 version tag, and are correctly interpreted by all three of inkscape 0.48/0.92 (read and write) and illustrator (read OK, write will convert everything to 72dpi pixels). The use-case for this are files created newly with Inkscape 0.92 and possibly edited with inkscape 0.48. Here's a sample file with the inkscape version tag removed: https://gist.githubusercontent.com/mgmax/eb73693a04da6d1b7a9f3d9681f12473/raw/9daf1932881e33392a20a6bc7d9f9e31761385c8/inkscape_0.92_without_version_tag.svg
- Files containing both inkscape and illustrator comments and only pixel-unit width are a huge problem. There is no unique way to know which one is right. The Inkscape extension should probably strip out the illustrator comment, and when VisiCut finds such a file it should display a serious warning. (That's why I don't like the warning messages occuring at almost every file)

Note: I assume that my findings for 0.48 are also valid for other pre-0.92 releases. "width" stands for the svg document root width/height.

Issue #375, opened 2016-07-01
Feature: Add command line arguments for provisioning


sedrubal:
Is it possible to add command line arguments to visicut for several provisioning tasks like:

 - download recommended VisiCut settings for a lab
 - install inkscape/Illustrator extension
 - ...

?

PR #415, opened 2017-05-13
Download settings: Pre-select lab based on FQDN

mgmax:
(currently only implemented for FAU FabLab)

If a PC has a matching hostname (FQDN), the "Download default settings" dialog defaults to the lab. This saves us some time because we have a separate account for each user, and we now lo longer have to tell every user to select the right lab.

Related to #375

### DIFF
--- a/src/com/t_oster/visicut/gui/MainView.java
+++ b/src/com/t_oster/visicut/gui/MainView.java
@@ -96,10 +96,12 @@ import java.io.InputStreamReader;
 import java.net.HttpURLConnection;
 import java.lang.reflect.Constructor;
 import java.lang.reflect.Method;
+import java.net.InetAddress;
 import java.net.URI;
 import java.net.URL;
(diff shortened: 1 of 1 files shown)


t-oster:
I think it is time to refactor that part into a config file, which contains a list of Labs together with URL and FQDN. Another if clause makes things ugly. If I find time next week, I will make a suggestion. However if you have time first, you are welcome to do it yourself.

Aditionally, we could make this list itself updateable, so we do not need to update VisiCut in order to add an URL.

mgmax:
I'm not sure whether this is worth the effort -- currently the whole list is just 15 easily understandable lines of code. In my opinion, we should focus on the "release blockers" for 1.8, but of course you are free to work on whatever you like :-)

mgmax:
Hm, and something is wrong with travis. Google suggests `sudo apt-get install ant-optional`.

PR #419, opened 2017-08-29
removed proprietary/non-free PngEncoder library

chca42:


mgmax:
Builds and works. @t-oster Please merge.

(I don't know where the documentation is, but the projector feature uploads a PNG image via POST to the specified projector URL. I extracted the image via Wireshark, and it is black background with red cutting lines for the objects, just as intended.)

PR #431, opened 2017-12-31
Do not warn about modern 96 DPI.

jnweiger:
Only warn about 72 DPI from Adobe Illustrator and 90 DPI from old inscape versions.
Files from inkscape 0.92 or later are silently(!) accepted as 96 DPI.
Files from unknown source are silently accepted as 90 DPI.
Other dubious 90 DPI sources should be added as theiy appear.


mgmax:
It would be even better to not warn about any file that contains a viewBox attribute.

jnweiger:
@mgmax Great idea. Thanks. Done.

PR #460, opened 2018-05-24
Add UI option to disable auto focus in main view, per job, instead of only in settings at a machine level.

hghile:


t-oster:
This is included in `1.8-27`. I had to fix a small error. Please test if it works correctly and if it destroys previous settings.

hghile:
Seems to be working, thanks!

Issue #458, opened 2018-05-06
UI scaling (vertical and horizontal) is buggy


mgmax:
No matter how large or small I resize the window, the scrollbars and UI resizing don't seem to make too much sense:

![screenshot large](https://user-images.githubusercontent.com/1520891/39677030-e1ccc554-5174-11e8-9097-aa00d6247404.png)

![screenshot small](https://user-images.githubusercontent.com/1520891/39677048-229fbe10-5175-11e8-8a3c-c7366024ca85.png)

The problem might to be caused by the scroll box around the mapping/position/lasersettings tab pane. This scrollbox stubbornly refuses to adapt to the available space.

@mrlinux1 I would like to remove the scrollbox which you added in commit f11597ba462b721ad2776e5876b5b64ff7c64140 , or find another solution to fix the layout for large enough monitors. Do you have any opinion or suggestions?

jnweiger:
![grafik](https://user-images.githubusercontent.com/1108546/39701595-112233ba-5202-11e8-984e-81c5526e29c9.png)

The intention of the scrollbox was to make it easier to reach the 'Execute' Button (Ausführen in German locale) in the lower right hand corner when screen space is limited. Currently the Scrollbar makes it possible to reach the Execute button without resizing. That is a good thing.

An enhancement would be (imho) to have a fixed footer with Calculate and Execute buttons always visible and a scrollbox for everything else above.

@mgmax Did you mean something different?

t-oster:
I did some changes to vertical resizing. Is this better: https://github.com/t-oster/VisiCut/pull/464

mgmax:
Now it's a lot better than before.

PR #464, opened 2018-05-28
Fix GUI resizing

t-oster:
Here are some changes done in the UI Editor in order to make resizing better.
Please test against the latest release in order to verify that this makes things better and not worse.
#458 
Here are some binaries:
http://download.visicut.org/feature-resizing


mgmax:
Thumbs up :+1: 

Looks way better:
Works for 1152x864 screen size (Win7).

and above...


As before, it does not work with 1024x768 screen size:


For comparison, before: 1024:
old-1024-768

before: 1152:
veryold-1152-768

t-oster:
Well, so I can merge this and we can improve it later...

PR #469, opened 2018-08-06
Improvement for Gui buttons in editableTablePanel

mrlinux1:
Moved the save and revert button up to the other buttons.

### DIFF
--- a/src/com/t_oster/uicomponents/EditableTablePanel.form
+++ b/src/com/t_oster/uicomponents/EditableTablePanel.form
@@ -45,14 +45,15 @@
               <Component id="btUp" min="-2" max="-2" attributes="0"/>
               <EmptySpace max="-2" attributes="0"/>
               <Component id="btDown" min="-2" max="-2" attributes="0"/>
-              <EmptySpace min="0" pref="0" max="-2" attributes="0"/>
+              <EmptySpace max="-2" attributes="0"/>
               <Component id="btRevert" min="-2" max="-2" attributes="0"/>
               <EmptySpace max="-2" attributes="0"/>
               <Component id="btLoad" min="-2" max="-2" attributes="0"/>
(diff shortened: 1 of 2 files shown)


t-oster:
thank you. But there is some space missing between the down and the revert button
bildschirmfoto vom 2018-08-07 11-37-14

I originally placed them on the bottom on purpose, because they are not working on a single item but on all items. However since the ui is so big and scrolling is necessary on many screens, it makes sense to move them up. Maybe we should still add some separator

jnweiger:
The spacer should be there. sure.
But for most of the panels, the '+' and all the additional buttons never made any sense to me. 
I am not sure, what a lasercutter would do, If we specify two rows. Is that for multiple laser guns?

But I also strongly agree, that the save button should not require using a scrollbar, before it can be found.
Alternative idea (mockup)
grafik

t-oster:
@jnweiger the lasercutter would cut your paths twice, possibly with different speeds/powers.

jnweiger:
Ah, thanks for the explanation! 

So we can make a multipass setting for cutting very thick material. 
It may be helpful for highly flamable materials (e.g. a stack of thin cardboard) to keep the laserhead moving fast with less power and make multiple passes instead.
And we can move the focus down a bit in every pass! Nice feature!
I have to explore that...

t-oster:
You are Welcome to improve the documentation ;)

mrlinux1:
Distance between Buttons fixed. All have the same distance now.

jnweiger:
https://github.com/t-oster/VisiCut/wiki/VisiCut-manual#lasersettings already has the explanation:
"You can add multiple treatments to one profile, which is useful for cutting very thick material in two steps."

Apparently I was never so desperate enough to read the docs, thanks for pointing there! :-)

PR #470, opened 2018-08-18
Remember if the window was maximized

mgmax:
and default to "maximized" if the previous setting is unknown or impossible (too large for the current screen).
Also correctly remember the size if the original window was placed outside the current screen coordinates (e.g. Linux two-screen setup, window on the second screen)

### DIFF
--- a/src/com/t_oster/visicut/gui/MainView.java
+++ b/src/com/t_oster/visicut/gui/MainView.java
@@ -407,12 +407,18 @@ public class MainView extends javax.swing.JFrame
       }
     });
     //apply the saved window size and position, if in current screen size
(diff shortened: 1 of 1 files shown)


t-oster:
Does this work If the window was on the second screen and now there is only one? Looks as If you just restore the bounds without moving. Or does the window manager take care of it?

mgmax:
As before, the position is ignored, I only set the size and don't move the window.

I assume moving would be too error-prone. For example, a maximized window has slightly negative x,y coordinates on Linux, so that Java thinks it's outside the screen bounds, and it seems that my second screen isn't even reported inside Java ...

PR #472, opened 2018-08-22
Engrave preview and multiple other improvements

mgmax:
This is an addition to my previous pullrequest (will rebase as soon as the other one is merged) with more discussion-worthy changes.

1. better antialiasing for the engrave preview by switching to a downscaled image if the user zooms out far enough
2. save some RAM by storing the (non-downscaled) raster preview as 1bit indexed (1 byte per pixel instead of 1 integer per pixel)
3. The heaviest change of this: switch from Thread.stop() to Thread.interrupt(), and dozens of interface changes to make the dithering interruptable and pass on InterruptedException. To be used together with the LibLaserCut pullrequest https://github.com/t-oster/LibLaserCut/pull/90 (compiles and works without, but the dithering will just go on without being interrupted). 

Please give feedback if this is the right approach and if you feel these are ready for merging. Testing is highly appreciated, I don't know how many stupid edge cases there are which I couldn't find. Mac is entirely untested, I will test Windows soon, but have no Mac. I suspect some graphics rendering stuff may be different on Mac.

Maybe a feature branch would be a good idea, so that others can easily test?

### DIFF
--- a/distribute/distribute.sh
+++ b/distribute/distribute.sh
@@ -1,5 +1,6 @@
 #!/bin/bash
 set -e # exit 1 if any command fails
+cd "$(dirname "$0")"
 # set -x # uncomment for debugging
 if [ "$1" == "--nocompile" ]
 then
@@ -7,8 +8,8 @@ then
 else
    COMPILE=1
 fi
-echo "Determining Version:"
-VERSION=$(cat ../src/com/t_oster/visicut/gui/resources/VisicutApp.properties |grep Application.version)
+echo "Determining Version (may be overriden with environment variable VERSION):"
+VERSION=${VERSION:-$(cat ../src/com/t_oster/visicut/gui/resources/VisicutApp.properties |grep Application.version)}
 VERSION=${VERSION#*=}
 VERSION=${VERSION// /}
 echo "Version is: \"$VERSION\""
--- a/distribute/files/VisiCut.Linux
+++ b/distribute/files/VisiCut.Linux
@@ -7,4 +7,4 @@ port=${DISPLAY##*:}
 port=${port%.*}
 # add it to the instance-port
 let port=port+6543
-java -Xms256m -Xmx1024m -jar "$DIR/Visicut.jar" --singleinstanceport $port "$@"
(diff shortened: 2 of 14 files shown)


mgmax:
Here is a collection of screenshots (please click and zoom in for full size) showing the effect of antialiasing downscaling, comparing Inkscape (left), VisiCut with new downscaling (center) and old VisiCut (right).

Antialiasing example

mgmax:
Things seem to work on Windows. (Note: actual lasercutter output is still untested, the dummy driver unfortunately doesn't support engraving)

t-oster:
Hi it seems as if VisiCut won't compile without this PR if I would pull the current master of LibLaserCut (with t-oster/LibLaserCut#90).

t-oster:
If you want to ship test-binaries, I can create a feature branch, so my buildserver will build it.

mgmax:
Yes, the change to LibLaserCut also required changes to VisiCut, because "throws" annotations were added, so it is incompatible with the old VisiCut code (but not the other way around: the new VisiCut code also works with the old LibLaserCut). 

If this situation lasts longer, I could create a commit which works around this by blindly catching these exceptions, but that would have to be reverted when the actual new code is merged. But that is extra effort and useless in the end.

Or you could revert the change in the main LibLaserCut code for now and just put it in an extra branch. Or leave LibLaserCut as it is and let the submodule refer to the old version.

t-oster:
I think I will just leave the old submodule version in VisiCut until you think your branch is ready to merge.

mgmax:
I have pushed some further commits to improve this, didn't find any bugs in my code. Feels like VisiCut 1.9.0, maybe a first release candidate?
Roughly tested on windows, and intensively tested (almost the same codebase, only with an upcoming new laser driver) on one Linux PC.

t-oster:
So this is ready to merge?

mgmax:
Yes, I would say it's ready to merge.

t-oster:
Ok. LibLaserCut is merged, so please update the submodule, rebase and squash what makes sense, then I can merge.

mgmax:
done, I updated liblasercommit only squashed two commits because I think the changes would become incomprehensible if squashed further (think of `git bisect`) and each commit makes sense for itself.

Issue #252, opened 2015-03-21
UTF8 characters in default mapping doesn't work under Win7


mgmax:
This bug is not so easy to reproduce. I suppose it is like this:

Loading the FAU FabLab settings, that were created under Linux, seems to work fine. (default mapping is "schneide rot ..." which contains an umlaut)
But manually setting the preferences (or even saving them unmodified) causes them to break.

Is there some charset issue?

I will test this further...

mgmax:
At least under WinServer2008 the umlaut is converted to a questionmark when saving the preferences. Under Linux this is not the case.

t-oster:
It may be, that filenames in Windows are saved in Windows-1252 instead of UTF8? IMHO in the java runtime, strings are always unicode, so I suggest that the file-names when written to disc are to blame?

PR #491, opened 2018-12-26
Fix problems with special characters in settings

mgmax:
Change all XStream serialization to UTF8.
Move all use of XStream.toXML, .fromXML to one single place to simplify future changes.

Should fix all problems with special characters (äöüß, ♥, etc). Tested with:
- Preferences -> Lab Name
- Preferences -> Default mapping (#252)
- material names
- Windows7 <-> Linux import/export
- import of old, existing settings on Linux (unchanged, here we already have UTF8) and Windows (works as good as possible, except where data is already broken*)

Not tested with Mac.

(* Some problems occur with *old* settings that cannot be fixed, because the special characters were replaced by '?' when saving with old VisiCut. However, the next time the preferences are saved everything is fine, and it seems that this concerns only unimportant details like "Lab name" and "Default mapping".)

Merry christmas everyone! :gift:


mgmax:
Upon further testing, I found a suspicious error message if the settings are already corrupted.
From this, I found out that the fallback to the old (pre-2012) XML format didn't work after my changes because the stream wasn't reset to the start or reopened. Now it should work, but I cannot find any test data for it. My best guess would be that we can remove the fallback code.

PR #516, opened 2019-08-02
Helpful error message if settings download fails

mgmax:


t-oster:
There seem to be some changes missing??

mgmax:
It's really just this one-line change so that the error message makes more sense. I now added a fix for travis-CI -- their oraclejdk8 seems broken.

PR #520, opened 2019-08-08
Add an attempt to use SSID to guess recommended lab settings

hghile:
wireless SSID seems like a good predictor for lab location.  I tested this on linux, but the mac one is just what the internet tells me should work.


t-oster:
Nice idea. Please use the Helper.isWindows etc functions to guess the OS and refactor your code into an Helper.getWifiSSID() function.

Maybe we should pull out the list of labs into a seprarate file and put the ssid and domain there to have it all on one place and encourage labs to not only add their name and url.

hghile:
Refactored.
Having lab settings not in code sounds like a good plan.  Having that be optionally paired with network info seems like it would work well too.

t-oster:
I did the other refactoring.

PR #534, opened 2019-10-17
Replace the obsolete com.apple.eawt with java.awt.desktop APIs. This …

yusufm:
…is required to run visicut on macs that have java > 8. And update travis CI config to swap jkd8 with jdk9.

See: http://openjdk.java.net/jeps/272 and https://docs.oracle.com/javase/9/migrate/toc.htm#JSMIG-GUID-C86F7403-7D7C-456F-8873-18C7F1BEDE2F

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: java
 jdk:
-  - openjdk8
+  - openjdk9
   - openjdk11
 
 before_install:
--- a/src/com/t_oster/visicut/gui/MainView.java
+++ b/src/com/t_oster/visicut/gui/MainView.java
@@ -22,11 +22,11 @@ package com.t_oster.visicut.gui;
 // It is loaded dynamically iff JavaFX is available:
 // import com.tur0kk.thingiverse.gui.ThingiverseLoginDialog;
 import com.tur0kk.thingiverse.gui.ThingiverseDialog;
-import com.apple.eawt.AppEvent.AboutEvent;
-import com.apple.eawt.AppEvent.OpenFilesEvent;
-import com.apple.eawt.AppEvent.PreferencesEvent;
-import com.apple.eawt.AppEvent.QuitEvent;
-import com.apple.eawt.QuitResponse;
+import java.awt.desktop.AboutEvent;
+import java.awt.desktop.OpenFilesEvent;
+import java.awt.desktop.PreferencesEvent;
+import java.awt.desktop.QuitEvent;
+import java.awt.desktop.QuitResponse;
 import com.frochr123.fabqr.FabQRFunctions;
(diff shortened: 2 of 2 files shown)


t-oster:
Is there a way to fix this without breaking Java 8 compatibility?

yusufm:
I’m looking into it, but I don’t think so. Not a simple way at least.

yusufm:
I can’t fine any simple way of maintaining both versions simultaneously.
Given that the java api has moved on a while ago, maybe it makes sense to
drop v8 support?

t-oster:
I think jre8 is widely used, so dropping it is too early IMHO.
However there is the option of a multi-release-jar-file: https://techbeacon.com/app-dev-testing/legacy-developers-guide-java-9

Theoretically it should be possible to use the java8 api and add classes with the same name which use the java9 api in the jar. Would make the build process more complicated though.
Maybe there is already some compatibility layer jar available, which we can just add to the classpath?

mgmax:
Why don't we just bundle Java 11 like we do for Windows? AdoptOpenJDK also has builds for Mac 64 bit.

t-oster:
Is a possible way, however we should not forget debian etc.

t-oster:
and while we're at it, shoudn't we use modularization and ship a custom (smaller) jre?

mgmax:
Does this change break Java 8 only on Mac, or everywhere? 

Regarding the situation on Debian and Ubuntu:

Debian stable already has Java 11. Same for Ubuntu 18.04. On Debian oldstable (end of life in one year or so), Java 11 is only available via the backports repository. The situation is probably similar for Ubuntu 16.04.

So, roughly summarized, requiring Java 11 will only hurt a few people now, and almost nobody in about one year.

t-oster:
well since java 11 is the new LTS it would be fine with me. No idea if it only breaks java 8 on osx, should be testable if you try to run the compiled jar with java8 on win/lin

t-oster:
it does break java 8 on all platforms.

yusufm:
I looked at the multi-release jar file, but it was adding complexity. Not sure if its something worthwhile to have long term in case things like this come up again.

Would it work to tag a version with java8 and just move master on to only support 9+?

mgmax:
I agree - let's move on to Java 11 and create an archive branch ("old-java8") for the last known version that works with java 8. I would suggest we use the slightly older version 095136a21602894e36a4cdb4574cde15d593f2e3 for the Java8 branch, because after that some changes were introduced only for fixing things on Java11.

That should be easier than the multi-release jar: If there actually are people who need Java8 and new features, they could backport (rebase) changes to the Java8 branch if they want to. If not (which I expect), then we haven't wasted any time.

PR #537, opened 2019-11-01
Make development great again

t-oster:
I am planning quite a few steps:

- [x] migrate liblasercut to Maven
- [ ] publish liblasercut to Maven Central
- [ ] remove swing application framework from all forms so they can be edited in stock netbeans
- [x] migrate visicut to Maven
- [x] fix distribute scripts etc


t-oster:
Hi @mgmax. I sucessfully migrated VisiCut to maven. All libraries are on central except for LibLaserCut (which I will of cause submit) and kabeja, which I will just include binary as it is now. However this is a major switch and will cause all sorts of problems with build scripts etc and will take some time until everything is consistent again. Also rebasing local branches may be a little harder because I have to rename the "com.t_oster" package to "de.thomas_oster", because I own that domain and I need to use it for submitting to maven central.

Since you are the most active member of the community I would like to know if you agree with these changes. IMHO this will make development better in the future and also allows for using LibLaserCut in other maven projects more easily.

mgmax:
I haven't used maven yet and haven't had the time to test this. Some random thoughts:

Removing the dependency JAR files from the git repository is good. It is also good that dependency versions are pinned so that builds are reproducible. 

Having LibLaserCut as a submodule made it rather easy to switch it out for a custom version, and also simplifies debugging and development. How will this look with maven? Is there a point in not including LibLaserCut as a submodule?

t-oster:
It would not be necessary to have LibLaserCut as a submodule. If you compile it on your machine, it will be installed in your local maven cache and automatically be used when building VisiCut. Also NetBeans recognized that the you have both open and even uses the complied-on-save classes from LibLaserCut in VisCut.

Technically we can even keep the submodule in order to reference specific commits and just change the build script to first build LibLaserCut and the VisiCut.

mgmax:
If I don't use netbeans but the commandline, can I still edit something in LibLaserCut, change to the VisiCut directory and `ant run`? Or is the workflow more complicated?

If it doesn't have too many disadvantages, I would suggest we keep LibLaserCut as a submodule for now - the move to maven is big enough.

mgmax:
I just tested the branch and get an error:
```
$ mvn install
Failed to execute goal on project visicut: Could not resolve dependencies for project de.thomas_oster:visicut:jar:1.9-SNAPSHOT: The following artifacts could not be resolved: org.kabeja:kabeja:jar:0.4, org.kabeja:kabeja-svg:jar:0.4, org.kabeja:kabeja-xslt:jar:0.4, de.thomas_oster:liblasercut:jar:1.9-SNAPSHOT: Could not find artifact org.kabeja:kabeja:jar:0.4 in central (https://repo.maven.apache.org/maven2)
```

t-oster:
No matter where you 'mvn jnstall' liblasercut, this version will be used for visicut. Try to build liblasercut first and maybe in visicut 'mvn prepare' is needed to install kabeja since it is not on Maven Central yet

t-oster:
Look at the .travis_ci file

mgmax:
That seems to work. My personal preference would be including LibLaserCut as submodule, and providing one-line replacements for `ant build` and `ant run`, e.g., move the travis_ci commands to `./build.sh`, and also provide a `./run.sh` to start VisiCut.

After this is done, we should have a look at making life easier for downstream (e.g. fablabnbg build their own version, and we at fau-fablab sometimes also do for emergency fixes). Currently, there are some things locked away in VisiCutBuilder which should IMO be moved to distribute.sh. Especially, generating the splash screen.

t-oster:
Ok git submodule makes sense to specify the exact commit anyway. The splash screen is in the makefile imho

mgmax:
Regarding "migrate away from Swing Application Framework": This code claims to provide a converter for migration: https://sourceforge.net/projects/nosaf/ .

t-oster:
@mgmax yes, I saw that but I did not get it to work yet. However I think there is some more work involved, because there is also no support for the beansbinding anymore.

Issue #492, opened 2018-12-26
consider listing features more prominently in Readme


matkoniecz:
https://github.com/t-oster/VisiCut#visicut - maybe screenshot or two to demonstrate that it is userfriendly?

First feature that is actually described is "upload and post a picture of a thing you made to Facebook", what - at least to me - is not something actually useful.

mgmax:
Thanks for pointing this out - I agree, and help with this would be appreciated. A list of important features could be copied from the website https://visicut.org (the sourcecode of the website is in the `gh-pages` branch). Adding screenshots to the website would be great -- and in my opinion the facebook/thingiverse stuff could be moved to the manual in the Wiki.

t-oster:
Yes, I am with you. Please help. Maybe we even remove the facebook and thingiverse stuff if no one needs it. On the other hand, as long as it does not cause issues, we can leave it there but it should not be listed as the first feature I agree

PR #539, opened 2019-11-07
remove info about social integration from readme

matkoniecz:
feel free to reject this PR, but at least for me this part of README is utterly useless and makes harder to find real information.

In my case I wanted to check whatever I can connect to a specific laser (ATMSolutions CO2 ATMS PRO1016 80 W - and I needed to wade through pointless Facebook integration stuff before discovering that this info is unavailable in README.

Facebook integration is anyway anti-feature, at least for me.

First step toward #492 by making space for listing features.

Again, feel free to close if you consider integration with social networks as an important feature - I am not actually user of this software (at this moment I am trying to find laser cutter controller for specific laser that will work on Linux)


t-oster:
I agree. However since people worked on the facebook and thingiverse integration and it' still in the code I would at least move the information to different files and add a reference to the readme instead of just removing it. 

However I also do not like the facebook and thingiverse integration in the main project and would like them to be in a plugin, but thats also a fair amount of work.

PR #541, opened 2019-11-09
Fix gui threadsafety

### LINKED ISSUE #538: Weird behavior on Windows since Java11
Since switching to Java11 (?), there are strange errors on Windows 10, especially in High-DPI setups:
Screenshot_20191103_145623
The screenshot is from my tests with a Virtualbox VM with 150% scaling enabled in the Windows display settings. This is not the only error (once I got one regarding TableModel, and once the preview did not react to changing the mapping), but the only one I could reliably reproduce. 

To reproduce, use Visicut with Fablab Berlin settings, open examples - basic - rectangle-20mm-green. Close Visicut.
Then it crashes about 2/10 times for the following: Open VisiCut, Files - recent files - rectangle-20mm-green.

My best guess would be that we have a huge threadsafety issue:
https://bugs.java.com/bugdatabase/view_bug.do?bug_id=6888157
https://www.tutorialspoint.com/is-swing-thread-safe-in-java
https://docs.oracle.com/javase/tutorial/uiswing/concurrency/index.html

mgmax:
This diagnoses and hopefully mostly fixes #538 . 

The crashes have a really low probability under Linux, but can be provoked by https://github.com/t-oster/VisiCut/commit/1c0b26bc78d65e9e511c41c1315daa76e9d9c3d0 .

In summary, we must never call GUI functions, e.g. `progressBar.setProgress()`, from any thread exept the Event Dispatcher Thread. This means that if we create a new thread, all GUI operations need to be encapsulated with `SwingUtilities.invokeAndWait` (or `invokeLater`). I wrote wrapper functions (ThreadUtils class) to make that easier: `assertInGUIThread` prints a warning if you are not in the right thread, and `runInGUIThread(()->{ progressBar.setProgress(42); });` calls `progressBar.setProgress(42);` in the GUI context (blocking until it finishes). The "weird syntax" `()->{...}` is a lambda expression; you can also use anonymous inner classes (subclass of Runnable as for Threads).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,6 +17,7 @@
 # Temporary files:
 /distribute/wintmp/
 *~
+*.orig
 
 # Maven stuff:
 /lib/nbproject/private/
--- a/src/com/t_oster/visicut/gui/ThreadUtils.java
+++ b/src/com/t_oster/visicut/gui/ThreadUtils.java
@@ -1,0 +1,76 @@
+/**
+ * This file is part of VisiCut.
+ * Copyright (C) 2011 - 2013 Thomas Oster <thomas.oster@rwth-aachen.de>
+ * RWTH Aachen University - 52062 Aachen, Germany
+ *
+ *     VisiCut is free software: you can redistribute it and/or modify
+ *     it under the terms of the GNU Lesser General Public License as published by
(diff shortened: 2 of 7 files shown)


mgmax:
Now rebased to current master

PR #545, opened 2019-11-24
distribute.sh: Show correct version when building without VisiCutBuilder

mgmax:
Before, when building with distribute.sh without VisiCutBuilder, the window title showed a fixed fallback string for the version, and the splash screen always showed "1.7".

Now, when building without VisiCutBuilder, the right version (determined from $VERSION or from git) is shown in the splash screen, and the window title simply doesn't show the version.

Nothing has changed when building *with* VisiCutBuilder, which writes the version into VisicutApp.properties.


t-oster:
Two questions:
- what happens if you do not use git ( download the zip from GitHub)?
- what happens when you run from netbeans?

mgmax:
The zip from GitHub doesn't work anyway since it doesn't contain the LibLaserCut submodule.

Netbeans does not start distribute.sh. It also does not use the splash image. The window title in Netbeans will simply be "Visicut" instead of "Visicut 1.7".

t-oster:
Ok, it would work if you previously download the liblasercut zip and mvn install it

mgmax:
I don't think so because distribute.sh calls make, which assumes that LibLaserCut is in the correct subdirectory.

PR #547, opened 2019-11-30
Split settings into public and private part

### LINKED ISSUE #546: Required files to import settings should not contain personnal data or volatile settings
When you want to share your devices, materials and lasercut settings, the file `setttings/settings.xml` is currently required, which was not the case previously.

Unfortunately, this file contains personal data and volatile settings.
This have two side effects:
 * `recentFiles` entry could be a personal data disclosure
 * `windowBounds`, `lastMaterial`, and maybe more, do appears in a `git diff` when you have your shared `visicut-settings` managed through VCS.

IMHO, all entries which are personal and/or volatile are user-defined and should be moved in a dedicated file. This way the new file should be explicitly excluded from import and a recommendation to ignore it in Git could be made in Wiki.

Note: This discussion follow the one started here:
https://github.com/haum/visicut-settings/issues/1#issuecomment-559681125

mgmax:
This more or less fixes #546.

It's not the ideal solution which would be splitting the fields into two disjunct files, <del>because that is currently impossible due to a dependency deadlock (new NetBeans cannot edit old GUI, old NetBeans cannot edit Java11 code). I hope @t-oster will manage to sort that out some time, </del>but for now I propose the following workaround:

There are two files:
- settings.private.xml: all settings including local transient settings.
- settings.xml: all settings, but with local transient settings nulled out

VisiCut tries to load settings.private.xml if it exists and is newer, otherwise falls back to settings.xml. VisiCut always saves both files. The local settings.private.xml may be deleted at any time.

Ignoring that file is currently the user's responsibility.

This change is backwards-compatible. It can be cleanly reverted without breaking anything relevant. The only downside is that if this change is reverted, the user will once loose his (rather worthless) temporary settings such as window size and recent files.

I have not tested this thoroughly but I am quite sure that the worst possible bug is that the transient settings may be lost, which one almost does not notice.
Proper testing and discussion is appreciated.

### DIFF
--- a/src/main/java/de/thomas_oster/visicut/Preferences.java
+++ b/src/main/java/de/thomas_oster/visicut/Preferences.java
@@ -33,6 +33,19 @@ public class Preferences
   {
   }
   
+  /**
+   * Clear all preferences which are user-specific, e.g., recent files and window-size
+   */
+  public void anonymize()
+  {
+    setLastMaterial(null);
+    setLastLaserDevice(null);
+    setRecentFiles(null);
+    setWindowBounds(null);
+    lastAutoUpdateTime=0;
+    setLastAutoUpdateLabName(null);
+  }
+
   private String defaultMapping = null;
 
   public String getDefaultMapping()
@@ -302,6 +315,10 @@ public class Preferences
    */
   public LinkedList<String> getRecentFiles()
   {
+    if (recentFiles == null)
+    {
+      recentFiles = new LinkedList<String>();
+    }
     return recentFiles;
   }
 
--- a/src/main/java/de/thomas_oster/visicut/managers/PreferencesManager.java
+++ b/src/main/java/de/thomas_oster/visicut/managers/PreferencesManager.java
@@ -62,11 +62,25 @@ public final class PreferencesManager
   }
   private Preferences preferences;
   
+  /**
+   * Get path of settings.xml, where all settings except some user-specific data (recent files, window size, etc.) are stored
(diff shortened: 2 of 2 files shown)


t-oster:
@mgmax UI editing should be possible with the Plugins mentioned in the Wiki (they work on new netbeans). I just want it to work with stock netbeans if possible any time soon.

mgmax:
@t-oster With NetBeans 11.2 I only get weird errors with the plugin. It worked with the old NetBeans 8.

mgmax:
I just saw that the wiki now mentions a second plugin -- I will try that.

mgmax:
@t-oster Yes, GUI editing seems to work again with both plugins as described in the wiki.
However, I don't have the time for the big ideal solution of splitting Preferences into two classes, adapting PreferencesManager, adding migration functions so that existing settings can be imported, and fixing the GUI and all other uses of preferences. 

Therefore, I still propose my workaround, which can be easily removed again without breaking the public settings.

t-oster:
I agree and I don't even see the necessity. I would make it even simpler so that if no settings.xml exist. It should just assume default values, but your solution is also fine with me.

mgmax:
At least for setting a sensible default mapping it is helpful to have settings.xml. Otherwise we could just skip it.

Would you suggest merging into master soon or would you prefer a public test branch (and build) for people to try out?

t-oster:
If you are certain enough that it won't break anything, just merge it.

PR #557, opened 2020-01-06
fix job name prefix 

jnweiger:
to not reset when clicking elsewhere and to be initialized from a shortened version of the loaded file name.
(Removing some old TODOs of mine).

### DIFF
--- a/src/main/java/de/thomas_oster/visicut/gui/MainView.java
+++ b/src/main/java/de/thomas_oster/visicut/gui/MainView.java
@@ -1712,6 +1712,26 @@ public class MainView extends javax.swing.JFrame
     {
       Logger.getLogger(MainView.class.getName()).log(Level.SEVERE, null, ex);
     }
+
+    if (!PreferencesManager.getInstance().getPreferences().isUseFilenamesForJobs())
+    {
+      // If we don't UseFilenamesForJobs implicitly, then we suggest a shortened job name in the prefix.
+      String jobName = file.getName();
+      if (jobName != null)
+      {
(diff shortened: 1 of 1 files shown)


src/main/java/de/thomas_oster/visicut/gui/MainView.java
```diff
@@ -2058,18 +2077,17 @@ private String getJobName()
     String jobname = "(unnamed job)";
     try
     {
+      // CAUTION: keep in sync with job name construction below in run()
```
mgmax: Is it possible to deduplicate the code? Then we could show the final job name (and not just the prefix) in the textbox. In my opinion, the label "Job name prefix" doesn't sound very user-friendly.

jnweiger: Not sure about the deduplication. Could break the threadding? or not?

the final job name has some parts that are not user editable. This can be an autoincremented Job Number or the full file name at the end. Depending on a users prefernce setting.

Showing the final job name next or below the prefix would be great. I don't dare to touch the XML for changing the GUI.

t-oster: You can change the UI from within Netbeans UI editor quite comfortable, you just need to have two plugins installed https://github.com/t-oster/VisiCut/wiki/Development:-Getting-started#editing-the-graphical-user-interface-layout

mgmax:
There's now a single point in the sourcecode where the job name is computed: `MainView.generateJobName()`.

t-oster:
@mgmax so this can be closed?

mgmax:
Closing due to a lack of response -- feel free to rewrite this PR for the new `MainView.generateJobName()`.

Issue #360, opened 2016-04-22
Cannot select laser cutter if starting with none and adding just one.


tealvince:
If the app is launched without the suggested settings (so no laser cutters are yet defined), then if one creates a single laser cutter afterwards, the laserDeviceComboBox is hidden while the selected laser device is still null.  This leaves no way to select a laser cutter or enable the "execute" button without creating a second laser cutter or restarting the program.

I'm not familiar with the code enough to know if there is a better solution, but this is what I did:

diff --git a/src/com/t_oster/visicut/gui/MainView.java b/src/com/t_oster/visicut/gui/MainView.java
index a642752..0dcbfc0 100644
--- a/src/com/t_oster/visicut/gui/MainView.java
+++ b/src/com/t_oster/visicut/gui/MainView.java
@@ -591,6 +591,7 @@ public class MainView extends javax.swing.JFrame
       this.laserCutterComboBox.setSelectedIndex(0);
       this.laserCutterComboBox.setVisible(false);
       this.jLabel9.setVisible(false);
-      this.visicutModel1.setSelectedLaserDevice(LaserDeviceManager.getInstance().getAll().get(0));
   }
   else
   {

pelrun:
I think it's better to move the `ignoreLaserCutterComboBoxUpdates=false` line above that if statement; the `this.laserCutterComboBox.setSelectedIndex(0)` line should kick off the appropriate update except the flag is suppressing it.

renebohne:
@tealvince  did you test what @pelrun suggested? Did it work and can you provide a pull request for this issue?

jnweiger:
The behaviour improved with newer versions of visicut:

The behaviour with only one laser now is:
 * The lasercutter is not shown. This is still confusing.
 * The execute button is activated without restarting.

Workaround: 
add a second dummy laser device, so that users see the exepcted image and name of their laser onscreen.

jnweiger:
btw. it is hard to see what the suggested solution is, from the provided snippet. All I can decipher is:

> +++ b/src/com/t_oster/visicut/gui/MainView.java
> @@ -591,6 +591,7 @@ public class MainView extends javax.swing.JFrame
> this.laserCutterComboBox.setSelectedIndex(0);
> this.laserCutterComboBox.setVisible(false);
> this.jLabel9.setVisible(false);
> 
>     * ```
>        this.visicutModel1.setSelectedLaserDevice(LaserDeviceManager.getInstance().getAll().get(0));
>       ```
>       
[...]

PR #559, opened 2020-01-06
trivial fix for https://github.com/t-oster/VisiCut/issues/360

jnweiger:
Caveat: ImageComboBox does not show the icon image, when isEnabled is false.
Keeping the ComboBox enabled all time thus would have a more consistent UX.
Choose one :-)


t-oster:
see #360 for the reason not to include this.

PR #564, opened 2020-02-17
Fix GUI Thread Warnings

t-oster:
@mgmax I am not so into the GUI Threading issue, but would it not be easy to just wrap all GUI modifying code in a
```
SwingUtilities.invokeLater(() -> {
  //do GUI stuff here
});
```
Block like this?

At least in https://github.com/t-oster/LibLaserCut/pull/136#issuecomment-581126598 @tatarize gets warnings because the driver uses the progresslistener, which modifies the GUI

### DIFF
--- a/src/main/java/de/thomas_oster/visicut/gui/MainView.java
+++ b/src/main/java/de/thomas_oster/visicut/gui/MainView.java
@@ -88,13 +88,9 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.io.InputStreamReader;
 import java.net.HttpURLConnection;
-import java.lang.reflect.Constructor;
-import java.lang.reflect.Method;
 import java.net.InetAddress;
-import java.net.URI;
 import java.net.URL;
 import java.net.URLConnection;
-import java.net.URLClassLoader;
 import java.net.UnknownHostException;
 import java.util.Arrays;
 import java.util.Collections;
@@ -119,6 +115,7 @@ import javax.swing.JMenuItem;
 import javax.swing.JOptionPane;
 import javax.swing.JScrollPane;
 import javax.swing.JViewport;
+import javax.swing.SwingUtilities;
 import javax.swing.filechooser.FileFilter;
 import org.jdesktop.application.Action;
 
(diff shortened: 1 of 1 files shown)


tatarize:
Yes. That is generally correct. Basically anything done in the driver is done in a different thread than the GUI thread. When you modify the GUI outside the GUI thread you should either get a hard-coded error or you get random bugs. The typical way to do correct this to check if the current thread is the gui thread, and if it is, run the requested operation. However, if the current thread is not the gui thread, to run then invokeLater() a run of this same command again. This tends to tie up loose ends since anything that is correctly being run in the GUI thread is run correctly, and outsources it to the gui thread when it isn't the gui thread.

You don't really want to create a bunch of run objects if things are being modified by the gui thread itself. So I wouldn't call *everything* in that. 

```java

import static javax.swing.SwingUtilities.invokeLater;
import static javax.swing.SwingUtilities.isEventDispatchThread;

    public void MyFunctionThatRequiresGUIThread(final float argument) {
        if (!isEventDispatchThread()) {
            invokeLater(() -> {
                MyFunctionThatRequiresGUIThread(argument);
            });
            return;
        }
        ModifyGuiInSomeWays();
    }
```

With the older java 7 or calls for new Runnable() or whatnot. Though if you're revisiting this, it might be nice to get something more generic than a progress listener that can do some stuff like issue warnings and give the user feedback on the USB being broken or some such thing. Or just a generic object that has more of a potential API to deal with things more broad than a progress listener.

mgmax:
In these trivial cases it's okay. Note that invokeLater may possibly hide exceptions, so for any nontrivial code that should be tested.

t-oster:
@mgmax well those are the cases where third-party code (i.e. liblasercut drivers) is executed which cannot ensure it is running on a GUI thread. @tatarize we could run the whole method in a gui thread, but I don't think this makes sense, because it would render the GUI non-responding during job sending, so if you @mgmax think it's OK, feel free to merge.

tatarize:
My suggestion there isn't to run in the gui thread, it's to check whether the calling thread is the gui thread and run it there if it is, but if it's not then toss off an invokeLater. The distinction there between Max's code and my suggested code there is if you ran the function in the GUI thread, it will create a runnable and invoke it later, when it doesn't need to do that since it's already in the correct thead.
```java
            public void taskChanged(final Object o, final String string)
            {
              if (!SwingUtilities.isEventDispatchThread()) {
                SwingUtilities.invokeLater(new Runnable() {
                  @Override
                  public void run()
                   {
                      taskChanged(o, string);
                    }
                });
                return;
              }
              MainView.this.progressBar.setString(string);
            }
```

You'd actually go through the main class and replace all instances of `ThreadUtils.assertInGUIThread();` with a variation of the above. Where if it is running in the gui thread, it checked one conditional and moved along, but if it's not, it turns around and runs it in the gui thread. Basically rather than sending off an error or warning you just recurse the call into the gui thread. It basically makes the entire class gui-thread-safe. Letting you call the functions from any thread and ensuring everything will work fine.

It also avoids code repeat, The function could be 1 line or 1000 and that little `call this in the gui thread if this isn't the gui thread` code fragment at the top remains the same 10 lines (of mostly scaffolding).

t-oster:
At least for the progress listener interface we can assume that we are not in the GUI thread because we call the send job method of the driver in a new non gui thread with in turn calls our progress listener methods, so I don't think a check is necessary

t-oster:
@mgmax I refactored the method a bit in order to reduce the calls to invokeLater. If you have no objections, I would merge this.

mgmax:
generateJobName() and getJobName() are highly similar, can we get rid of the latter?

t-oster:
We can refactor that later

Issue #565, opened 2020-02-19
Release VisiCut as AppImage for Linux


TheAssassin:
Providing an [AppImage](https://appimage.org/) would have, among others, these advantages:

- Applications packaged as an AppImage can run on many distributions (including Ubuntu, Fedora, openSUSE, CentOS, elementaryOS, Linux Mint, and others)
- One app = one file = super simple for users: just download one AppImage file, [make it executable](https://docs.appimage.org/user-guide/run-appimages.html#download-make-executable-run), and run
- No unpacking or installation necessary
- No root needed
- No system libraries changed
- Works out of the box, no installation of runtimes needed
- Optional [desktop integration](https://docs.appimage.org/user-guide/run-appimages.html#integrating-appimages-into-the-desktop)
- Optional binary delta updates, e.g., for continuous builds (only download the binary diff) using [AppImageUpdate](https://docs.appimage.org/introduction/software-overview.html#appimageupdate)
- Can optionally sign your AppImages (inside the file) using gpg2
- Works on Live ISOs
- Can use the same AppImages when dual-booting multiple distributions
- Can be listed in the [AppImageHub](https://appimage.github.io/apps) central directory of available AppImages
- Can double as a self-extracting compressed archive with the [`--appimage-extract`](https://docs.appimage.org/user-guide/run-appimages.html#mount-or-extract-appimages) parameter

[Here is an overview](https://appimage.github.io/apps) of projects that are already distributing upstream-provided, official AppImages.

t-oster:
Feel free to provide a build script, preferably dockerized

TheAssassin:
Dockerizing this shouldn't be a big deal. The biggest problem will be to bundle everything for Java 11.
I can just put it in `distribute/linux`, I guess?

t-oster:
Yes. If the docker file downloads everything we don't need to bloat the repo with Java binaries etc

TheAssassin:
Mission accomplished, just have to test the result on a few different distros. I've used [appimagecraft](https://github.com/TheAssassin/appimagecraft) to eliminate most boilerplate code. Unfortunately the existing Makefile stuff in this repo doesn't seem to be able to build out-of-source, but we can optimize that later.

I intend to bundle Inkscape and VisiCut together with our laser configs in a single AppImage that is then distributed on all our workstations (as well as visitors' computers). Bundling VisiCut was amazingly simple, since you provide those `SNAPSHOT` JARs which are mostly self-contained.

[...]

PR #566, opened 2020-02-19
Add AppImage build scripts

TheAssassin:
Contributes to #565.

This PR adds support for [appimagecraft](https://github.com/TheAssassin/appimagecraft). To build AppImages, download appimagecraft (you can either install it via pip or just use the AppImages from its release page), and run it in the repository root directory.

As requested, I also added a Docker-based build script.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -24,3 +24,7 @@
 /lib/nbproject/private/
 /lib/build/
 /target/
+
+# AppImage builds
+.appimagecraft-build*/
+*.AppImage
--- a/appimagecraft.yml
+++ b/appimagecraft.yml
@@ -1,0 +1,34 @@
+version: 1
+
+project:
+  name: org.visicut.visicut
+  version_command: git describe --tags
+
+build:
+  null:
+
+scripts:
+  post_build:
+    - set -x
+    # TODO: check if out-of-source builds are possible
+    - pushd "$PROJECT_ROOT"
+    - make
+    - popd
+    - |2
+      mkdir -p AppDir/usr/{bin,jre}
+      wget -O- https://github.com/AdoptOpenJDK/openjdk11-binaries/releases/download/jdk-11.0.6%2B10/OpenJDK11U-jre_x64_linux_hotspot_11.0.6_10.tar.gz | \
+          tar xz --strip-components=1 -C AppDir/usr/jre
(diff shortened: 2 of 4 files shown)


t-oster:
Looks good to me, thanks. When I have time I will evaluate the possibility for automatic builds and publishing

TheAssassin:
The classic approach for making AppImages is usually to let Travis CI build and upload binaries for every commit, tag, ... The advantage is that you always have an easy-to-test binary for every build which you can hand out to people. LibreOffice for instance has also started to make AppImages over a year ago, and being able to ship beta releases to testers and translators was part of the motivation.

mgmax:
Can the download of appimagecraft be pinned to some specific hash? I'm not so happy about blindly downloading and running whatever is the newest in some external repository.

The appimage should include all files included in the "platform independent zip" built by distribute.sh, as otherwise some functionality will be broken (e.g., installing the inkscape extension, default example settings). Preferably, use the zip file as input so that we don't have to duplicate the logic. (The "interface" of distribute.sh may be problematic, feel free to change this to commandline arguments, with a fallback to querying interactively if nothing is specified. We then need to update VisiCutBuilder to use the new interface; I can help with that.)

TheAssassin:
> Can the download of appimagecraft be pinned to some specific hash? I'm not so happy about blindly downloading and running whatever is the newest in some external repository.

Yes, if you install it with `pip` you can specify a commit ID. Most appimage tools are however "rolling release", and don't introduce breaking changes. Versioning is not available for most dependencies of appimagecraft. Within all the years I've been part of the AppImage project, this concept has worked out well.

> The appimage should include all files included in the "platform independent zip" built by distribute.sh, as otherwise some functionality will be broken (e.g., installing the inkscape extension, default example settings). Preferably, use the zip file as input so that we don't have to duplicate the logic. (The "interface" of distribute.sh may be problematic, feel free to change this to commandline arguments, with a fallback to querying interactively if nothing is specified. We then need to update VisiCutBuilder to use the new interface; I can help with that.)

That's entirely possible. The appimagecraft script can run any command required to build that zip, and then extract it into the AppDir which is subsequently turned into an AppImage. I don't see any urge to do so, though; so far I have not found anything missing in the VisiCut AppImage I've built two weeks ago.

Perhaps you should re-post your comment in #565 so it won't be forgotten?

PR #569, opened 2020-03-07
Make order/dithering options more accessible

mgmax:
I'd like to make the dithering and vector sorting options more discoverable by showing them besides the speed/power settings. Currently I am stuck with how to make editing work -- I can show an edit dialog but it seems impossible to apply the modified profile, at least without major changes to the code structure.

screenshot

### DIFF
--- a/src/main/java/de/thomas_oster/visicut/gui/propertypanel/PropertyPanel.form
+++ b/src/main/java/de/thomas_oster/visicut/gui/propertypanel/PropertyPanel.form
@@ -17,12 +17,17 @@
     <DimensionLayout dim="0">
       <Group type="103" groupAlignment="0" attributes="0">
           <Group type="102" attributes="0">
-              <EmptySpace min="-2" max="-2" attributes="0"/>
+              <EmptySpace max="-2" attributes="0"/>
               <Group type="103" groupAlignment="0" attributes="0">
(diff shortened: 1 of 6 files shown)


tatarize:
Yeah, I ran into similar problems in MeerK40t. There's a lot of issues with the order of operations for the various operations. There's a series of things you need for a proper raster and you really kinda need them done in the correct order. The only thing I came up with is to flag the block with a series of operations that are not applied yet, and let them be manually applied later.

So you'd flag an image to be resampled, curve corrected, and dithered. And you'd need those operations to happen in that order. You can't dither then resize and you wouldn't want to color correct before the resample. But, you could well see how the image looks and want to change the original curve corrections later, which you can't really do if you've actually already modified the image.

The solution I finally settled on was doing something like modifiers in Blender, or adjustment layers in photoshop. Basically you queue up a series of operations, for an image and just delay the execution. Then you can mandate the order to be correct. 

This sort of stuff would almost always require pretty large code modifications. But, some of the stuff it would allow makes it kinda worth it.

https://github.com/meerk40t/meerk40t/issues/151

I'm pretty sure it's a non-starter. Though I totally recommend moving those dithering algorithms out of the driver. They don't and have never belonged there. Get some proper image manipulation code.

mgmax:
I'm giving up on this for now. Better focus on keeping what we already have in a stable condition.

PR #572, opened 2020-03-14
Fix #571: Show progress while saving file

### LINKED ISSUE #571: File export completed indicator
This is an issue our laser cutter users have reported when using Visicut.

Essentially ( on Linux, all versions I have tried, including most recent ), there is no indicator telling the user when the file has been succesfully exported ( File -> Export laser code ).

The user goes to export the laser code, if it's a pretty simple file, that happens instantly, and if the file is very complex ( say it's a bitmap engraving ) it takes some time. But the user can't know if the file has been exported and they can move on to use it, or if the file is still processing, and they need to wait.

I'm not sure what the correct way of handling this would be. Maybe have an information message display in the same area warnings and errors get displayed, that tells user the file was succesfully exported?

Any improvement of this would be very welcome by our users. I'm available to test anything that gets changed regarding this. Thanks !

mgmax:
Now Visicut will show exactly the same progress bar and message for "Export Laser Code" as it does for "Execute Job".

@t-oster This will cause a minor merge conflict with #564 - should I wait for that?

### DIFF
--- a/src/main/java/de/thomas_oster/visicut/gui/MainView.java
+++ b/src/main/java/de/thomas_oster/visicut/gui/MainView.java
@@ -1959,6 +1959,7 @@ public class MainView extends javax.swing.JFrame
       }
     }
     this.jTextFieldJobName.setText("");
+    this.exportGcodeMenuItem.setEnabled(execute);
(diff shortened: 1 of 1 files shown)


t-oster:
No :)

PR #580, opened 2020-05-24
Cleanup

phdoerfler:
Mostly Javadoc related spring cleaning :)


src/main/java/de/thomas_oster/visicut/gui/EditVectorProfileDialog.java
```diff
@@ -145,8 +145,8 @@ private void initComponents() {
         cbPPI.setEditable(true);
 
         binding = org.jdesktop.beansbinding.Bindings.createAutoBinding(org.jdesktop.beansbinding.AutoBinding.UpdateStrategy.READ_WRITE, this, org.jdesktop.beansbinding.ELProperty.create("${vectorProfile.DPI}"), cbPPI, org.jdesktop.beansbinding.BeanProperty.create("selectedItem"), "ppi");
-        binding.setSourceNullValue(new Double(0));
-        binding.setSourceUnreadableValue(new Double(0));
+        binding.setSourceNullValue(0d);
+        binding.setSourceUnreadableValue(0d);
```
mgmax, line 149: this code is auto-generated from the NetBeans UI file `EditVectorProfileDialog.form` (see `<editor-fold ...` above), any change will be reverted as soon as you save the UI again. I suggest keeping it as it is.
```suggestion
        binding.setSourceNullValue(new Double(0));
        binding.setSourceUnreadableValue(new Double(0));
```

phdoerfler: Uh, right. The joys of auto generated code. Yeah, makes not much sense to mess around with that.

PR #581, opened 2020-05-25
Fix Travis CI build

mgmax:
@t-oster I'm not sure if your latest commit ("WIP") to master was intentional. Here's a possible fix so that travis works again, though it's somewhat wasteful to build all the packages even if there is no release.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,7 @@
-language: java
+language: minimal
 os: linux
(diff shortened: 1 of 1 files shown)


t-oster:
Yes I was trying to do a normal build and only use the docker container for tagged commits. Then I wanted to reuse the maven cache inside the container but I couldn't get it to work

PR #598, opened 2021-01-21
fixed typos

freddii:


resources
```diff
@@ -0,0 +1,2 @@
+LICENSE:6:- The Camera Icon src/com/t_oster/visicut/gui/ressources/camera.png is from the Token Icon Set of EVAN BROOKS (http://brsev.com/). Its licensed under "Creative Commons Attribution Non-commercial (by-nc)"
```
mgmax, line 1: looks like you accidentally added a new file?

mgmax (changes requested):

resources
freddii: Yes that was done by mistake.

freddii:
fixed it in #599

Issue #502, opened 2019-03-24
Empty "Exemple" folder in the MacOS French version


waumand:
Hi makers !
With last Java 8 (version 201) with MacOS Mojave and 3 different Mac, in French, the "Exemple" folder is empty.

I can't use exemples to make some shapes :/

Any idea ?

<img width="1440" alt="Capture d’écran 2019-03-24 à 17 41 47" src="https://user-images.githubusercontent.com/6934787/54882832-659a1580-4e5e-11e9-8b28-3917da137a91.png">
<img width="488" alt="Capture d’écran 2019-03-24 à 17 43 07" src="https://user-images.githubusercontent.com/6934787/54882833-659a1580-4e5e-11e9-918e-a17997b8fea7.png">

t-oster:
Wich settings did you download?

waumand:
@t-oster when you say setting you speak about setting file ?
I use this one for my Fablab with the RS-1060L 3D printer
[visicut-1060.vcsettings.zip](https://github.com/t-oster/VisiCut/files/3000818/visicut-1060.vcsettings.zip)
http://robotseed.com/index.php?id_category=11&controller=category&id_lang=1

t-oster:
I can confirm this. The examples included in the specific downloaded settings work, but the predefined examples are not found. Probably a path issue. I won't have time in the next two week though, but after that I will look into it.

PR #620, opened 2021-09-20
Macos inkscape extension install fix

### LINKED ISSUE #586: Inkscape extensions not included in OSX-Package
The wiki says "In VisiCut, select Extras -> Install Inkscape extension... to install the extension to Inkscape. Then you can restart Inkscape, open a file, Extensions -> Lasercut -> Open with VisiCut." which I have done before but in the 1.9-70-g2c0db535 version (at least the Mac version), there isn't an Extras menu and the Actions menu is empty.  How do I install the extension?

sebastian-schmid:
This Pull Request will fix #586 and #502 

Would be also great if someone can test on windows to make sure this is not breaking something.


sebastian-schmid:
A build from this code works on MacOS 10.15 with Inkscape 1.1. Also tested the deb package on Linux Mint, which is still working.

I think it should also still work on Windows. Can someone please test that just to make sure it does not break anything on windows? (I don't have any Windows System)

Would also appriciate feedback to the commited code.

t-oster:
Hi, I am not quite sure if the VisiCut.MacOS script is the right place. IMHO the inkscape extension should try to find the VisiCut.app bundle here https://github.com/t-oster/VisiCut/blob/master/tools/inkscape_extension/visicut_export.py#L262 and only if this is not present it should look for the VisiCut.MacOS script.
The VisiCut.MacOS script should only be present in the platform independent-zip distribution but not in the .app bundle.

distribute/files/VisiCut.MacOS
```diff
@@ -1,7 +1,10 @@
 #!/bin/bash
-DIR="$( cd -P "$( dirname "$0" )" && pwd )"
-
-# Print and check java version
-java -version 2>&1 | tee /dev/stderr | egrep -q '"[1-9][0-9]\..*"' || echo "Warning: Your Java version may be too old. We recommend Java 11 or newer. Trying anyway."
-
-java -Xms256m -Xmx2048m -jar "$DIR/Visicut.jar" "$@"
+bundle_identifier="$(osascript -e 'id of app "VisiCut"')"
+if [ $? -eq 0 ]
```
mgmax, line 3: If both the app and the ZIP file version are present, then the ZIP version would always start the app. I don't think that is intended.

mgmax:
It is certainly a matter of taste where we put this logic. My suggestion was to keep the extension as simple as possible, so that we don't have to copy many changes to each extension for Inkscape, Illustrator and whatever comes in the future.

jnweiger:
Tested on win10  v21H1 with inkscape 1.1

Works perfectly fine. Thank you!

distribute/files/VisiCut.MacOS
sebastian-schmid: I think its the opposite way. The MacApp Version is preferred over the ZIP version if an MacApp is found. I will try to change this in a way that it will always use the version from which the extension is installed.

t-oster:
So is this ready to merge?

sebastian-schmid:
No, i am still working on a solution that will not prefer the installed version over the zip version if both are "installed" on the system.

mgmax:
Actually, there is a variable like that:
https://github.com/t-oster/VisiCut/blob/9e819d34b501d015cc9fe7bc214ca3516db410b7/src/main/java/de/thomas_oster/visicut/misc/Helper.java#L292-L311

IMO, the issue is rather in `VisiCut.MacOS`, which should *first* look for `(dirname $0)/Visicut.jar` and if that fails, open the App bundle. Currently, the order is reversed.

My suggestion for `VisiCut.MacOS` is roughly:
```
if  test -e $DIR/Visicut.jar; then
   java -jar ...
else
   open -b "$bundle_identifier" --args "$@"
fi
```

t-oster:
But where is VisiCut.MacOS present on a system where only the app bundle is installed?

sebastian-schmid:
Changed it, now it first checks if Visicut.jar is in the PATH, if not checks if MacApp is installed and then starts it else echo an simple error.

Please review :)

t-oster:
I thought the visicut.macos script is only present in the zip distribution, not in the OSX bundle.

sebastian-schmid:
Yes, this File is present in the MacOS App Bundle :+1:

Issue #633, opened 2022-01-14
Fix position file X=0 Y=0


LaFabrique35:
Hello, when I import an SVG file into VisiCut, I would like it to automatically position itself at X = 0 and Y = 0 on the work surface. The settings are correctly configured in the options but the file is not positioned at the correct coordinates.

t-oster:
How is your SVG view box? Usually the view box is positioned 0/0 so you can absolutely position elements in the svg

LaFabrique35:
My SVG view box is positioned with an origin at the bottom left. Ditto on Inkscape. How to force the file to be positioned in X 0 Y 0 during the import?

t-oster:
So it is top left an you're expecting bottom left? We should make this corresponding to the Lasercutter settings. Until then just make the SVG view box fill the laser bed

PR #635, opened 2022-01-19
Make importers aware of bottom left coordinate system

t-oster:
In #633 @LaFabrique35 suggests, that if the lasercutter has a bottom-left coordinate system, the files should be placed relative to the bottom left.
E.g. if you have this SVG:

[example.zip](https://github.com/t-oster/VisiCut/files/7900304/example.zip)

the rectangle is positioned x=20 and y=30 (relative to the bottom of the view box).

I started implementing this, but I have a few things to consider:

1. Is this what all users would expect? The SVG file itself has the coordinate system always top left as dictated by the SVG standard, but the position where the SVG viewbox is placed on the laserbed is changed.
2. What about the G-Code importer? Should we interpret all Y-Coordinates in the G-Code as relative to bottom left (is anyone using the G-Code importer anyway?)
3. Should the Raster-Importers (JPG,PNG etc) import the drawing also bottom-left?

Binaries are at https://download.visicut.org/feature-633-import-bottom-left

### DIFF
--- a/src/main/java/de/thomas_oster/visicut/VisicutModel.java
+++ b/src/main/java/de/thomas_oster/visicut/VisicutModel.java
@@ -699,7 +699,7 @@ public class VisicutModel extends Component // FIXME: "extends Component" isn't 
 
   public PlfPart loadGraphicFile(File f, List<String> warnings) throws ImportException
   {
-    return this.getGraphicFileImporter().importFile(f, warnings);
+    return this.getGraphicFileImporter().importFile(f, getSelectedLaserDevice().isOriginBottomLeft(), getSelectedLaserDevice().getLaserCutter().getBedHeight(), warnings);
   }
 
 
--- a/src/main/java/de/thomas_oster/visicut/model/graphicelements/AbstractImporter.java
+++ b/src/main/java/de/thomas_oster/visicut/model/graphicelements/AbstractImporter.java
@@ -30,14 +30,14 @@ import java.util.List;
 public abstract class AbstractImporter implements Importer
 {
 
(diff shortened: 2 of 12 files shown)


LaFabrique35:
Wow, what responsiveness! This new proposed version works very well! I was able to test it with the users of my FabLab, it works really well! Thank you !

mgmax:
Sorry for the long delay, I was busy with other stuff :-)
It is certainly good for the one original user request, but I'm not sure about how this fits as a general solution.

Looking at the code: My feeling is that we're slightly mixing up the interface boundaries. In an ideal solution, we would separate between
- "import the file and remember the page size" (convert SVG/JPG/whatever to PlfPart. Store the page size in some property, if the format has some definition of page size. GCode does not. JPG is always "page size = image size".)
- "move it to the bottom corner" (PlfPart.moveToBottomLeft() ?)
- decide whether to move to the bottom corner (separate checkbox for preferred position ?)
- configure where the zero point in the GUI ruler should be.

At least it would be nice to keep this y-shifting out of the individual image importers, or to the very least rename the method arguments from "originIsBottomLeft" to "placePageAtBottom"

Regarding functionality, I'm still struggling to fully understand the use case. I don't think that the position of the origin is crucial. The origin is artificial and doesn't matter for how you use the lasercutter. Maybe the position of the physical "ruler" where you physically align your rectangular piece of material is more relevant: If your cutter only has that on bottom-left, then (without VisiCam camera view) you will prefer to push your material against that corner for repeatable positioning.

mgmax (changes requested):
Needs discussion, see my previous comment from a long time ago

PR #639, opened 2022-02-08
WIP: flatpak distribution

t-oster:
This is a work in progress. I am working on a flatpak distribution, but I am not yet sure how to integrate the build process....


mgmax:
I like the idea; not sure how it will cooperate with the Inkscape extension. At least the other way around (Inkscape from flatpack/snap/... calling out to another program) does not work as far as I remember. We may have to use some workaround (socket file, polling in some directory, etc.).

TheAssassin:
This branch has become very much stale. There is a portable AppImage deployment available. Maybe it should be closed.

PR #641, opened 2022-02-22
Hghile visicut lab

hghile:
Add a new link for SLU makerspace settings.

Took me a bit to find the new file location.. would be nice to update the example link in https://github.com/t-oster/VisiCut/wiki/How-to-add-default-settings-for-your-lab


t-oster:
Thanks for the PR. Feel free to update the Wiki

PR #653, opened 2022-09-27
Bug fixes for Inkscape extension and Mac support

khenriks:
This PR contains 4 separate commits, with different kinds of fixes (see commit descriptions). I'm happy to split into multiple PRs if desired. In brief:

1. Fix logic for opening Visicut socket in Inkscape extension. The previous logic doesn't work with Python 3 or properly handle errors.
2. Fix and simplify logic for launching Inkscape in extension. The previous logic doesn't work with Python 3 or properly handle errors.
3. Fix Mac script to launch Visicut.
4. Fix Mac icons.


mgmax:
Someone should check if the code still works fine with Python2. (Necessary for older Inkscape, e.g. 0.92.x on Windows).
Then I'm happy to merge.

khenriks:
I tested it with Python 2.7, and unfortunately the `os.fsencode` function does not exist. I updated the code to handle this case, and the rest works fine. Please let me know if this looks okay.

Issue #670, opened 2023-05-24
Set up CI/CD completely with GitHub actions


TheAssassin:
To make development easier and allow contributors to test their changes in the real world before they send PRs, I would like to move CI/CD completely to GitHub actions. This would further allow for continuously deploying VisiCut in the form of a moving tag for which a prerelease can be created (this is what many AppImages do for instance, and the AppImage community provides tools that make this very easy).

In case you feel uncomfortable sharing secrets with GitHub (e.g., for code signing), you could set up a custom runner or perform the final steps on other private infrastructure. Both GitHub actions and any other CI/CD system should share the same scripts.

mgmax:
I agree. Feel free to do so. 🙂 https://github.com/t-oster/VisicutBuilder should be a good starting point for the build environment/Docker container.

The visicut repo already contains "distribute.sh" which does 95% of all the building. It only lacks a few details such as the bundled Windows EXE for autotrace. These remaining 5% should be moved over from VisicutBuilder.

TheAssassin:
I already rewrote most of `distribute.sh` (just didn't get to creating a PR) to make it more usable and include the AppImage build in it.

TheAssassin:
#671 introduced a prototype Docker based build system based on the old AppImage-specific Dockerfile. Ideally, this will replace other Docker images to avoid redundancies.

mgmax:
related: https://github.com/t-oster/VisiCut/pull/680

TheAssassin:
PR available in #710.

Issue #621, opened 2021-09-21
Inkscape extension doesn't connect with visicut


Jorgininho:
Hello,

I'm working on a fresh install of inkscape and visicut, but the export from inkscape to visicut via extension menu doesn't seems to be working.
Here is the configuration : 
OS : Ubuntu 20.04.3 LTS
Inkscape : 1.1
Visicut : 1.9-102

When I try to export a design from inkscape to visicut, i get this : 
```
Traceback (most recent call last):
  File "visicut_export.py", line 267, in <module>
    VISICUTBIN = which("VisiCut.Linux", [VISICUTDIR, "/usr/share/visicut"])
  File "visicut_export.py", line 118, in which
    raise Exception("Cannot find executable {0} in PATH={1}.\n\n"
Exception: Cannot find executable 'VisiCut.Linux' in PATH=['/usr/share/visicut', '/usr/share/visicut', '/snap/inkscape/9090/bin', '/snap/inkscape/9090/usr/sbin', '/snap/inkscape/9090/usr/bin', '/snap/inkscape/9090/sbin', '/snap/inkscape/9090/bin', '/usr/local/sbin', '/usr/local/bin', '/usr/sbin', '/usr/bin', '/sbin', '/bin', '/usr/games', '/usr/local/games', ''].

Please report this bug on https://github.com/t-oster/VisiCut/issues

For a quick fix: Set VISICUTDIR and INKSCAPEDIR in /home/cervin/.config/inkscape/extensions/visicut_export.py

```

I don't see how to apply the quick fix solution, since the executable are indeed present in the mentioned folders

Thanks

mgmax:
My guess is that you have installed Inkscape as a "snap package". As far as I understand, snap packages are very locked down and can not call outside applications. As far as I understand, you even require special permission to call one snap application from another, and there is no way to launch non-snap applications -- not the best design...

The extension should at least be fixed to include a warning if "/snap/..." is on the PATH and Visicut can not be found.

You can try two workarounds:
- Start Visicut and then call the extension. If Visicut is already running, it will be contacted via network, maybe this is allowed in snap.
- Uninstall inkscape and re-install it from the normal "deb" packages (for example, using `sudo apt install inkscape` in the terminal window). However, you will then probably have an older Inkscape version.

mgmax:
A newer VisiCut version is available via https://download.visicut.org/ .

[...]

PR #671, opened 2023-05-27
Improve deployment and fix AppImage

TheAssassin:
Contributes to #565, #670.

This pull request rewrites `distribute.sh` to make it easier to use, introduce building in temporary directory (to isolate build processes from one another), allows users to choose which distribution formats to build for using CLI parameters, and a lot more improvements. It also adds AppImage as an output format by integrating the previous [appimagecraft](https://github.com/TheAssassin/appimagecraft) setup). The AppImage also now contains a full `visicut` directory like the other setups, which means the extensions can be installed.

Along with the new/updated script,  a Docker build script is shipped that allows for building all of the target distribution formats on any platform. The image uses Debian as a base, which ships all the necessary tools (NSIS, `makepkg`, `checkinstall` etc.).

The Arch Linux stuff doesn't seem to build in the new Dockerized setup, but it's up to someone else to test and fix that.

In my opinion, this project should focus on portable formats like AppImage and drop support for hacks like `checkinstall` or distribution-specific formats. Third parties can maintain distribution-specific formats, either upstream or as part of user repositories (AUR, Launchpad, ...). Let's rather make the AppImage a flawless user experience, so issues like #621 won't ever have to be worked on. The less complexity, the better.

The Inkscape plugin probably needs quite a bit more love (there's lots of old Python 2 style code, and overall it seems a bit messy), but it works now when installed from an AppImage, which is a great improvement.


TheAssassin:
This automatic "review" is mostly nonsense. I just moved code around. I do not intend to fix less-than-optimal code from before in this PR.

t-oster:
Looks good to me at first glance. You forgot to **chmod +x** the distribute-docker.sh, but that's a minor.
In order to build the arch-package (makepkg) on ubuntu there is a whole lot of stuff involved (see https://gitlab.com/t-oster/visicutbuildservice/-/blob/master/Dockerfile), but since most arch users would build from the AUR anyway instead of downloading binary packages, we should probably just ditch the archlinux build and add an automated upload of the PKGBUILD to the AUR later.

TheAssassin:
Pretty much my thoughts as well. I'll drop them in a commit. We can rebase before merging.

I'm not sure how popular the `.deb`s are, but honestly, the current `checkinstall` setup is a bad hack. It works to some extent, but I think binaries should be built by the distributions if any. How about dropping them as well?

t-oster:
Also your version detection won't work because it's not in the project directory. We should sourround it with pushd/popd

t-oster:
> Pretty much my thoughts as well. I'll drop them in a commit. We can rebase before merging.
> 
> I'm not sure how popular the `.deb`s are, but honestly, the current `checkinstall` setup is a bad hack. It works to some extent, but I think binaries should be built by the distributions if any. How about dropping them as well?

Not sure. I don't think we are popular enough that any debian/ubuntu distro will include visicut, so people will need packages. I am not sure how good app-images feel to fresh-installed ubuntu users. My goal was to make visicut downloadable and runnable for the average user of any popular OS in the most standard way. I did not yet try to download and run the appimage on a fresh ubuntu VM, but if that feels smooth enough, we could drop it.
But as long as it works, who cares if it's a hack?

t-oster:
Ok. Now I have to look at my automated build server. I guess there are some things to fix with the new distribute script, but today is too late/early.

Thanks for your contribution.

TheAssassin:
Well, that was unexpected.

> I don't think we are popular enough that any debian/ubuntu distro will include visicut, so people will need packages. 

Just need to ask, I guess. If the packaging process was be smoother, I think they wouldn't mind. Debian for instance has got some experience with shipping Java-based packages.

Anyway, I'm certainly biased, being a maintainer of the AppImage project and author of many tools around it. We should discuss those things in an issue. Feel free to ping me if you need input.

distribute/Dockerfile
```diff
@@ -0,0 +1,23 @@
...
+
+ENV DEBIAN_FRONTEND=noninteractive
+
+# note: don't build with OpenJDK > 16, as this is the runtime we ship
+RUN apt-get update && \
+    apt-get install -y openjdk-11-jdk git wget shellcheck bash make maven python3-minimal python3-pip \
+        nsis zip libfuse2 checkinstall makepkg && \
+    python3 -m pip install git+https://github.com/TheAssassin/appimagecraft.git#egg=appimagecraft
+
+# note: makepkg is contained in universe only, so we need to add that repository
```
mgmax, line 15: I don't understand what this comment is supposed to mean here. No "universe" repository is added.

distribute/distribute.sh
```diff
@@ -1,193 +1,288 @@
...
+if [[ "${1:-}" == "" ]]; then
+    log "Usage: [env BUILD=1] $0 <targets>"
+    echo
+    log "Available targets:"
+    log "    - zip"
+    log "    - windows-nsis"
+    log "    - macos-bundle"
+    log "    - linux-appimage"
+    log "    - linux-checkinstall"
+    log "    - linux-makepkg"
```
mgmax, line 29: there is no such target in the below script

distribute/distribute.sh
```diff
@@ -1,193 +1,288 @@
...
+        local url="$1"
+        local hash="$2"
+
+        pushd "$build_dir"
+
+        wget --content-disposition "$url"
+        downloaded_file="$(find . -iname '*.zip' -or -iname '*.tar*' -type f | head -n1)"
+        echo "$hash  $downloaded_file" | sha256sum -c
+
+        # -bb1 displays a list of extracted files
```
mgmax, line 119: I don't understand the context of this comment

distribute/Dockerfile
```diff
@@ -0,0 +1,23 @@
+# makepkg is available in bullseye's backports (and newer distros) only
+# checkinstall is available in bullseye only, too
+FROM debian:bullseye-backports
+
+SHELL ["bash", "-x", "-c"]
+
+ENV DEBIAN_FRONTEND=noninteractive
+
+# note: don't build with OpenJDK > 16, as this is the runtime we ship
```
mgmax, line 9: Our DEB package indicates that it is also compatible with Java 11:
`--requires 'bash,openjdk-11-jre\|openjdk-17-jre,potrace' \`

Java supports backwards-compatible compilation, depending on some setting in the maven file. https://maven.apache.org/plugins/maven-compiler-plugin/examples/set-compiler-release.html

distribute/distribute.sh
```diff
@@ -1,193 +1,288 @@
...
+    log "Available targets:"
+    log "    - zip"
+    log "    - windows-nsis"
+    log "    - macos-bundle"
+    log "    - linux-appimage"
+    log "    - linux-checkinstall"
+    log "    - linux-makepkg"
+    echo
+    log "Available environment variables:"
+    log "    - \$BUILD=[...]: if set to any string, $0 will build VisiCut's JAR first (not set by default)"
```
mgmax, line 32: I'm not so happy about not building by default. The docker-build.sh also builds by default, and the previous distribute.sh also did it.

It simplifies usability if there is just one "do everything" script that builds a complete package by default.

src/main/java/de/thomas_oster/visicut/misc/Helper.java
```diff
@@ -300,6 +305,10 @@ else if (isMacOS()) {
         String line = r.readLine();
         while (line != null)
         {
+          if (appImagePath != null && line.equals("VISICUTBIN = \"\"")) {
+            System.err.println("APPIMAGE=" + appImagePath);
```
mgmax, line 309: debug output should use `Logger`.

tools/inkscape_extension/visicut_export.py
```diff
@@ -71,14 +69,16 @@ def get_single_instance_port():
                 sys.stderr.write("Warning: Cannot determine session ID. please report this.\n")
     return port
 
-# if Visicut or Inkscape cannot be found, change these lines here to VISICUTDIR="C:/Programs/Visicut" or wherever you installed it.
+# if Visicut or Inkscape cannot be found in PATH, change these lines here to VISICUTDIR="C:/Programs/Visicut" or wherever you installed it.
```
mgmax, line 72: The intended meaning was: "If the extension tells you that it can not find VisiCut or Inkscape, change these lines to where your VisiCut or Inkscape is." The target audience does not necessarily know what "PATH" is.

tools/inkscape_extension/visicut_export.py
```diff
@@ -336,21 +337,38 @@ def get_original_filename(filename):
 
 # Try to start own VisiCut instance
 try:
-    creationflags = 0
-    close_fds = False
     if os.name == "nt":
         DETACHED_PROCESS = 8  # start as "daemon"
         creationflags = DETACHED_PROCESS
         close_fds = True
```
mgmax, line 343: these lines can be removed in your proposed version

tools/inkscape_extension/visicut_export.py
```diff
@@ -336,21 +337,38 @@ def get_original_filename(filename):
...
-        try:
-            import daemonize
-            daemonize.createDaemon()
-        except:
-            sys.stderr.write("Could not daemonize. Sorry, but Inkscape was blocked until VisiCut is closed")
-    cmd = [VISICUTBIN] + arguments + [dest_filename]
-    Popen(cmd, creationflags=creationflags, close_fds=close_fds)
-except:
-    sys.stderr.write("Can not start VisiCut (" + str(sys.exc_info()[0]) + "). Please start manually or change the VISICUTDIR variable in the Inkscape-Extension script\n")
+        # on POSIX-compliant OSes (Linux, macOS etc.) we can detach from the parent process using the classic double
```
mgmax, line 349: Your change removes the need for `daemonize.py`. Therefore that file should be dropped.
https://github.com/t-oster/VisiCut/blob/master/tools/inkscape_extension/daemonize.py

tools/inkscape_extension/visicut_export.py
```diff
@@ -336,21 +337,38 @@ def get_original_filename(filename):
...
+            [VISICUTBIN, *arguments, dest_filename],
+            cwd="/",
+            start_new_session=True,
+            stdout=subprocess.DEVNULL,
+            stderr=subprocess.DEVNULL
+        )
+
+except Exception as e:
+    print(
+        f"Can not start VisiCut ({repr(e)}). Please start manually or change the VISICUTDIR or VISICUTBINvariables in"
```
mgmax, line 368: missing space after "VISICUTBIN"

distribute/distribute.sh
```diff
@@ -1,193 +1,288 @@
...
+for target in "$@"; do
+    log "Building for target $target"
+
+    build_dir="$(mktemp -d -t visicut-build-$target-XXXXXX)"
+    cleanup() {
+        [[ -d "$build_dir" ]] && rm -rf "${build_dir:?}"/
+    }
+    trap cleanup EXIT
+
+    download_and_extract_jdk() {
```
mgmax, line 109: I would suggest moving this function outside the for loop since it doesn't depend on the loop variables.

mgmax (commented):
Although this was already merged I still have some suggestions and comments. Glad to see that things are going forward!

mgmax:
Regarding the need for a DEB package: As far as I know, it is the cleanest and best integrated installation method we can offer on Debian/Ubuntu. No need to bundle the JRE, which depends on processor architecture, `glibc` versions et cetera. Automatic integration into the start menu. etc.

Inkscape still has a [whole bunch of issues](https://gitlab.com/groups/inkscape/-/issues/?sort=created_date&state=opened&label_name%5B%5D=packaging%3A%3AAppImage&last_page_size=20&page_before=eyJjcmVhdGVkX2F0IjoiMjAxOS0wOS0yNCAxNjowMzo0NC4xNDg4NzcwMDAgKzAwMDAiLCJpZCI6IjI1MjUzMTMyIn0) with AppImage, so I expect that we will also have some issues and need the DEB package as a reliable fallback option.

TheAssassin:
> Although this was already merged

This may have been merged prematurely.

> Inkscape still has a whole bunch of issues with AppImage

None of which affect *third party tools* packaged as AppImages themselves. As said, the Inkscape extension is working just fine with the AppImage now.

> As far as I know, it is the cleanest and best integrated installation method we can offer on Debian/Ubuntu. No need to bundle the JRE, which depends on processor architecture, glibc versions et cetera. Automatic integration into the start menu. etc.

All the experience I have collected since I joined AppImage in 2017 (and having been a full-time open-source developer for years now, too), I can tell you that this is neither clean nor that bundling a JRE is a problem. I'd rather build different kinds of AppImages (which is really easy, too, the JRE bundles provide AArch64, too). Not every institution who want to use VisiCut on Linux use Debian or derivatives. For instance, two labs I know prefer openSUSE resp. Fedora.

But let's discuss the AppImage vs. native packaging question in a new issue. I'm sure that, if you try an AppImage built by the new set of scripts, you'll like what you see.

I'll send a follow-up pull request. I might make `distribute.sh` a Python script, though, so that the help text can be derived from the argument parser object.

distribute/Dockerfile
TheAssassin: Left over from the times I used Ubuntu as a base.

distribute/distribute.sh
TheAssassin: I used to use 7-zip, which works fine for `.zip`, but for `.tar.*` you need to call it twice.

distribute/Dockerfile
TheAssassin: Please feel free to change this. I don't see a reason not to use OpenJDK 11 at this point, as it is still a supported release as far as I can see.

distribute/distribute.sh
TheAssassin: We can change the default and have users set a `NO_BUILD=1` or something like that.

src/main/java/de/thomas_oster/visicut/misc/Helper.java
TheAssassin: This is an oversight, I'll get rid of it.

tools/inkscape_extension/visicut_export.py
TheAssassin: I see. I thought this was a third-party dependency to `pip install`.

PR #685, opened 2023-09-06
Don't assume hard-coded bash location. Replace `which` with `command`.

hzeller:
Two little changes to make shell scripts work more universally.

Not all Unixes have bash installed in /bin/bash; use the /usr/bin/env bash pattern to make finding it more platform independent.

The `which` command is deprecated in some Linux distributions (e.g. Debian); replacing it with more canonical command -v.


TheAssassin (approved):

t-oster:
Thanks for your contribution

PR #686, opened 2023-09-12
Fix location of embedded JRE in macOS bundle

### LINKED ISSUE #684: VisiCutMac-1.9-200-g5624a69b JAVA error
I downloaded version 1.9-200 from https://download.visicut.org/, but I'm unable to open it on my mac. I get the error:
```
ERROR launching 'VisiCut'.

No suitable Java version found on your system!
This program requires Java 11 or later.
Make sure you install the required Java version.
```

I have Java 20:
```
java --version
openjdk 20.0.2 2023-07-18
OpenJDK Runtime Environment Zulu20.32+11-CA (build 20.0.2+9)
OpenJDK 64-Bit Server VM Zulu20.32+11-CA (build 20.0.2+9, mixed mode, sharing)
```

and my `JAVA_HOME` is set correctly. In fact, [Version 1.9.1](https://github.com/t-oster/VisiCut/releases/tag/1.9-71) works just fine. It looks like the newest 1.9-200 version is reading the java version incorrectly

khenriks:
This was inadvertently broken in 4bd60b0. The JRE is intended to be installed under Contents/Plugins/JRE, but the refactoring inadvertently made it be installed under Contents/Plugins only. In turn, this meant the launched script could no longer find the JRE and gave an error instead of starting the application.

Fixes #684.

### DIFF
--- a/distribute/distribute.sh
+++ b/distribute/distribute.sh
@@ -228,7 +228,7 @@ EOF
 
             # deploy jre
             mkdir -p VisiCut.app/Contents/Plugins/
-            mv jre/* VisiCut.app/Contents/Plugins/
+            mv jre/ VisiCut.app/Contents/Plugins/JRE/
 
             # create bundle
             zip -r bundle.zip VisiCut.app/


t-oster:
Thanks for the quick fix

PR #687, opened 2023-09-27
Suggest to preserve the user entered text  in the JobName TextField

jnweiger:
Currently visicut deletes the value, when a) the laser is changed, b) material is changed, or c) the mapping is changed.
This makes it tedious, if I want to e.g. laser the same file different mappings, which I frequently do. I'd prefer to just edit the prefix, instead of entering it again in full each time.
Preserving the text is also nicer in the case an inexperienced user wants to enter the job name before choosing a mapping.

Also, I suggest to insert a whitespace after the prefix to make it look nicer.

### DIFF
--- a/src/main/java/de/thomas_oster/visicut/gui/MainView.java
+++ b/src/main/java/de/thomas_oster/visicut/gui/MainView.java
@@ -1986,7 +1986,6 @@ public class MainView extends javax.swing.JFrame
         execute = false;
       }
     }
-    this.jTextFieldJobName.setText("");
     this.exportGcodeMenuItem.setEnabled(execute);
     this.calculateTimeButton.setEnabled(execute);
     this.executeJobButton.setEnabled(execute);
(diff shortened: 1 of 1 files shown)


ToWipf:
That would be really helpful, it happens to me very often that I have to re-enter the text

jnweiger:
@MaxGaukler @t-oster do you have an opinion about this?

mgmax:
I don't have time for a full review; here are some first guesses from only looking at the code.

It's not optimal to never reset the field, but certainly better than resetting it way too often. It might, however, be confusing if user A uses VisiCut, sets a name, and then user B goes to the same PC, loads a file into VisiCut and then inherits the same job name. In total it is still better than before.

Looking at the diff: You did something unrelated to LibLaserCut, please revert. Otherwise it should be OK.

Some unrelated notes:
- I have no clue if anyone still uses the QR code editing stuff, I guess we should remove that feature entirely
- Personally I'm not really happy with the "job name prefix" field in general: UX could be better - people don't know what "prefix" means, and it doesn't show that there is already an automatic default value even if you enter nothing.

jnweiger:
I've manually restored commit 1a5e5c49c17bd862b63f2234b02edcbd6153cace into the LibLasercut submodule, that seems to help to get rid of the unintended change there. 

@t-oster What do you think about resetting the jobname less frequently?

t-oster:
Hi, I am not actively using visicut at the moment, so if @mgmax agrees, I am fine with it.

PR #692, opened 2023-11-19
Remove Projector / FabQR feature

mgmax:
The feature for QR code scanning and projector output was no longer usable because it depended on a really old PHP script for age-old PHP versions.

Still needs some practical testing: In an earlier version, sending laser jobs was completely broken. Now there might be some other issues remaining.


TheAssassin (approved):
Had a look and built a version with the suggested changes. Looks good to me. I think it's a brilliant idea to remove features that have not been used in practice since their original implementation (in this case, 9 years ago as part of some thesis).

PR #697, opened 2023-12-21
Enhancement: larger MappingBox

jnweiger:
This PR improves usability of the MappingBox by saving a few mouse miles:
 - whenever a file is loaded into visicut, the user has to choose a mapping. As this is a repetitive task, the clicks and mouse-movements should be as little as needed.
 - with custom defined mappings, (which I frequently use) the user has to 1) click open the combo box, 2) move the mouse into the widget (or its scrollbar). 3) scroll down (using mouse wheel or drag). 4) click to choose the mapping.

This PR simply extends the scrollbox to 20 elements,
- we we have plenty of empty space below, it usually does not obscure anything
- in case of a very small screen, the scrollbox automatically opens upwards, going ca 350 pixels high. This should always fit on the screen.
- user interaction is reduced to two clicks: open, choose.

Original size:
grafik

----

New appearance:
grafik

----

New appearance, when there is little space below:
grafik


mgmax:
Thanks :-)

You can also set a default mapping in the options, so users normally don't have to choose.

PR #721, opened 2024-04-08
Fix Visicut calling inkscape in an AppImage.

jnweiger:
In Ubuntu 20.04, this worked flawlessly. But only by chance. We use libraries from the host system, together with the binary inside the mounted appimage. in Ubintu 22.04, this crashes:

/tmp/.mount_inkscaYGPL8G/usr/bin/inkscape --version
 /tmp/.mount_inkscaYGPL8G/usr/bin/inkscape: error while loading shared libraries: libboost_filesystem.so.1.71.0: cannot open shared object file: No such file or directory

The fix is to not call the binary directly, but the AppRun script, which nicely prepares the environment. But then, AppRun also prints a message for the user to stdout, (instead of stderr). We now need to filter this, when parsing output from e.g. --version:

/tmp/.mount_inkscaYGPL8G/AppRun --version 2>/dev/null
 You should not use AppImage in production, but you can speedup the AppImage by following this guide: https://inkscape.org/learn/appimage/
 Inkscape 1.3.2 (091e20e, 2023-11-25)


jnweiger:
This PR obsoletes https://wiki.fablab-nuernberg.de/index.php?title=VisiCut&action=edit&section=2

tools/inkscape_extension/visicut_export.py
```diff
@@ -141,12 +141,19 @@ def is_exe(fpath):
 def inkscape_version():
     """Return Inkscape version number as float, e.g. version "0.92.4" --> return: float 0.92"""
     version = subprocess.check_output([INKSCAPEBIN, "--version"],  stderr=DEVNULL).decode('ASCII', 'ignore')
+    if not version.startswith("Inkscape "):
```
TheAssassin, line 144: No need for this check, you can run this line-wise "grep" every time.

tools/inkscape_extension/visicut_export.py
```diff
@@ -326,6 +333,17 @@ def get_original_filename(filename):
...
+# We detect this by checking for an AppRun file, in one of the parent folders of our INKSCAPEBIN.
+# If so, replace INKSCAPEBIN with AppRun, as this is the only safe way to call inkscape.
+# (a direct call mixes libraries from the host system with the appimage, may or may not work.)
+dir = os.path.split(INKSCAPEBIN)[0]
+while dir != '/':
+    if os.path.exists(os.path.join(dir, "AppRun")):
+        INKSCAPEBIN = os.path.join(dir, "AppRun")
+        break
+    dir = os.path.split(dir)[0]
+
```
TheAssassin, line 346:
```suggestion
dir = os.path.dirname(INKSCAPEBIN)
while dir != '/':
    apprun_path = os.path.join(dir, "AppRun")
    if os.path.exists(apprun_path):
        INKSCAPEBIN = apprun_path
        break
    dir = os.path.dirname(dir)
```

What is the compatibility target for this code? Python 3.4 introduced `pathlib` which makes such code a bit more convenient.

TheAssassin: Please also bear in mind that this code will make the extension work *only* for the currently running Inkscape. The following approach should make it work permanently:

```
appimage = os.environ.get("APPIMAGE")
if appimage and os.path.exists(appimage):
    INKSCAPE_BIN = appimage
else:
    # code from above
```

The reason to include the other code is that the AppImage _might_ be extracted by some users to a static location. Yes, that's very much annoying and yes, they're technically not AppImages any more. But that's what some people do. Personally, I don't care as much, I'm happy with the three-liner already.

TheAssassin (changes requested):

tools/inkscape_extension/visicut_export.py
jnweiger: In principal, I'd be happy to keep compatibility with all python versions, currently supported. 
As we support a very broad range of platforms, I don't want to make assumptions what python might be running in all our linux, windows or mac environments.

Apparently we already use array unpacking in `subprocess.Popen(
            [VISICUTBIN, *arguments, dest_filename]` which was introduced in python 3.5(?) :see_no_evil: 
This probably already limits backwards compatibility in an unintended way.

TheAssassin: Then again, Python 3.4 and 3.5 were released in 2014 and 2015 respectively hand have come EOL in 2019 and 2020 respectively. There's no point in keeping compatibility to dead software.

The oldest distro that might still use Python 3.4 is CentOS 7 which is due for EOL in June, i.e., not even 2 months from now.

tools/inkscape_extension/visicut_export.py
jnweiger: I am just defensive here. By testing the assert condition already here, I am sure, that whatever my code breaks, it will only break in cases that would have failed before anyway.

tools/inkscape_extension/visicut_export.py
```diff
@@ -384,3 +402,4 @@ def get_original_filename(filename):
     sys.exit(1)
 
 # TODO (complicated, probably WONTFIX): cleanup temporary directories -- this is really difficult because we need to make sure that visicut no longer needs the file, even for reloading!
+# - maybe add the PID od the running visicut, then we can detect orphaned temp direcories.
```
jnweiger, line 405:
```suggestion
# - Maybe add the PID od the running visicut, then we can detect orphaned temp direcories.
```

Silly text change to retrigger CI.

tools/inkscape_extension/visicut_export.py
jnweiger: Thus, the compatibility target question is answerd :-)

jnweiger: Afaik, the environment variable APPIMAGE points to the appimage file, rather than to anything inside a mounted or unpacked appimage.
I've chosen the latter, as this is noticeably more efficient.

I don't understand your comment about "the extension work only for the currently running Inkscape" -- isn't that the whole point of an extension to run with the currently running inkscape? 
I cant imagine a real world scenario, where the original inkscape was closed before the extension would make its cleanup call. 
(That might exist as a race condition, if the user starts the extension and then quickly kills inkscape. But what should the user expect in that case anyway?)

tools/inkscape_extension/visicut_export.py
TheAssassin: You can just re-run failed actions directly, though.

TheAssassin:
I'll open a separate PR to fix the issues in the linux-checkinstall CI.

tools/inkscape_extension/visicut_export.py
mgmax: I would prefer a simpler version that always does the same. The currently proposed code is a little confusing because in one branch, version "leaks" from the for loop, while in the other branch, it is the original version variable.

Maybe something like
```
version_raw = subprocess....(...)
# get first line of stdout that starts with "Inkscape"
lines = version_raw.splitlines()
version = [line for line in lines if line.startsWith("Inkscape ")]
assert length(version) == 1, "inkscape --version did not return version number"
```

mgmax:
Please also report the stdout pollution as an upstream issue with Inkscape. We are probably not the only one affected by that.

( Side topics, not really relevant to this PR: What about the other combinations and directions - does calling Visicut from the AppImage Inkscape work correctly, even with all the AppImage environment variables? What if VisiCut is also an AppImage? )

TheAssassin:
*If* someone opens an issue upstream, please ask them to just `export INKSCAPE_VERSION=1234` for plugins.

tools/inkscape_extension/visicut_export.py
jnweiger: Done that. Please re-review.

jnweiger:
https://gitlab.com/inkscape/inbox/-/issues/10199

jnweiger:
> ( Side topics, not really relevant to this PR: What about the other combinations and directions - does calling Visicut from the AppImage Inkscape work correctly, ...

Yes, I believe most real-life combinations work. I have especially tested the case where both, Inkscape and VisiCut are AppImages.

tools/inkscape_extension/visicut_export.py
```diff
@@ -326,6 +330,17 @@ def get_original_filename(filename):
         VISICUTBIN = which("VisiCut.Linux", [VISICUTDIR, "/usr/share/visicut"])
 INKSCAPEBIN = which("inkscape", [INKSCAPEDIR])
 
+## Test if this inkscape is in an appimage.
+# We detect this by checking for an AppRun file, in one of the parent folders of our INKSCAPEBIN.
+# If so, replace INKSCAPEBIN with AppRun, as this is the only safe way to call inkscape.
+# (a direct call mixes libraries from the host system with the appimage, may or may not work.)
+dir = os.path.dirname(INKSCAPEBIN)
+while dir != '/':
```
mgmax, line 338: On Windows this gives an infinite loop because `dir` ends up being something else, e.g., `C:\`

mgmax (changes requested):

TheAssassin:
> does calling Visicut from the AppImage Inkscape work correctly, ...

This is broken currently and will be working once this has been merged.

tools/inkscape_extension/visicut_export.py
TheAssassin: Indeed. In fact, there are better (and thus safer!) APIs available nowadays. Python's pathlib is well prepared for such a scenario:

```
from pathlib import Path

dir = Path(INKSCAPEBIN).parent
assert dir.is_dir()

# one could also use other metrics as an abort condition, but this will even catch the FUSE mountpoint
while not dir.is_mount():
    apprun = dir / "AppRun"
    if apprun.is_file() and apprun.is_executable():
        INKSCAPEBIN = apprun_path
        break
    dir = dir.parent

# it would be even easier with the parents property (an iterator)
# example:
# >>> list(Path("/ab/cd/ef/bin/visicut").parents)
# [PosixPath('/ab/cd/ef/bin'), PosixPath('/ab/cd/ef'), PosixPath('/ab/cd'), PosixPath('/ab'), PosixPath('/')
for parent in Path(INSCAPEBIN).parents:
    apprun = parent / "AppRun"
    if apprun.is_file() and apprun.is_executable():
        INKSCAPEBIN = apprun_path
        break

Comment (jnweiger) [reply tools/inkscape_extension/visicut_export.py:338]: Indeed, did not think about windows. Thanks for catching this!
I like the very readable .parents iterator!

Comment (TheAssassin) [reply tools/inkscape_extension/visicut_export.py:338]: Indeed, it is very handy. I was about to write my own but then remembered to have a look into pathlib first.

File: tools/inkscape_extension/visicut_export.py [hunk 5]
```diff
@@ -334,13 +335,12 @@ def get_original_filename(filename):
...
 # If so, replace INKSCAPEBIN with AppRun, as this is the only safe way to call inkscape.
 # (a direct call mixes libraries from the host system with the appimage, may or may not work.)
-dir = os.path.dirname(INKSCAPEBIN)
-while dir != '/':
-    apprun_path = os.path.join(dir, "AppRun")
-    if os.path.exists(apprun_path):
-        INKSCAPEBIN = apprun_path
+for parent in Path(INKSCAPEBIN).parents:
+    apprun = parent / "AppRun"
+    if apprun.is_file() and os.access(apprun, os.X_OK):
```
TheAssassin, line 340: I like this `os.access` check, it's very compact. `pathlib` should have a shortcut for that.

jnweiger: Maybe only checking is_file() alone is even more compact? An inkscape AppImage that has an 'AppRun' file which has no execute bits, or is mounted without exec, or suffering from a security policy, or whatever ... is out of scope here anyway.

TheAssassin: Just failing silently is not the best idea. Your algorithm can still cross mountpoints (that's what the `is_mount` check was there for before) and I can imagine a bunch of false positive detections of `AppRun`s in that case. For instance, in some occasions, AppDirs can be integrated in each other (lazy approach but it works, I use that myself occasionally). In such a case, the execute bit on some `AppRun` in the chain might not be set. Also, AppImages can be extracted and copied to devices that do not have the executable bit set, etc.

The check is fine and it's good to have it. In fact, security wise, we should add the mountpoint check and abort once it's crossed. To do so, this should be appended to the `for`'s body:

```python
    if parent.is_mount():
        break
```

Security (i.e., preventing critical issues) and safety (i.e., keeping users from experiencing weird behavior) always add a bit of verbosity to the code, but in the end, it's always worth it. We'd have some good reasons to throw exceptions here even. But that'd bloat this PR. I think in its current form, it's a reasonably solid implementation and I don't see any attack vector (without thinking about it for more than 5 minutes).

mgmax: I found an `INKSCAPE_COMMAND` environment variable that seems to point to the AppRun file. Can you check if we can use this instead? First guess:
```
INKSCAPEBIN = os.environ.get("INKSCAPE_COMMAND") or which("inkscape", [INKSCAPEDIR])
```
This would avoid the potential problems.

mgmax (changes requested):

TheAssassin:
What does "found" mean? Is that an officially supported environment variable provided by the Inkscape code? Is it an AppImage-specific feature?

mgmax:
For Inkscape 1.3.x and 1.4.x (current / upcoming release), the AppImage sets `INKSCAPE_COMMAND`.
https://gitlab.com/inkscape/inkscape/-/blob/1.4.x/packaging/appimage/AppRun?ref_type=heads#L40-44
This was added in 2022: https://gitlab.com/inkscape/inkscape/-/merge_requests/4751

For Inkscape 1.5.x (development branch) the AppImage is currently broken and fails to launch Python, so it is not possible to test. https://gitlab.com/inkscape/inkscape/-/issues/4851

Inkscape's `inkex` library for writing extensions has been using it since at least 1.0, so I guess that the normal Inkscape installation >= 1.0 also sets the variable.
https://gitlab.com/inkscape/extensions/-/blob/master/inkex/command.py?ref_type=heads#L46

Once we drop support for AppImage-Inkscape < 1.3 and normal Inkscape < 1.0 (used e.g. on Ubuntu 20.04) we could throw out half of our extension code and instead use `inkex.command` and friends. Until then, I suggest to do it as above: Use `INKSCAPE_COMMAND` if available, else fall back to what we did before.

tools/inkscape_extension/visicut_export.py
mgmax: @jnweiger Could you please change to use `INKSCAPE_COMMAND` (and check if it works), so that we can merge?

mgmax: ping :-)

jnweiger: I no longer remember this issue. Sorry. What should we do?
First check if INKSCAPE_COMMAND exists, if not try INKSCAPE_BIN, if not try to find AppRun with one of the loops discussed here?
Or just use INKSCAPE_COMMAND and fail if it fails?

I cannot comment on issues caused by "AppDirs integrated in each other" -- if an is_mount() check helps, then please add that where appropriate.

I am happy to test drive a PR that incorporates all these wishes and concerns that we collected here.
But writing that PR is beyond me. Sorry.

mgmax: My suggestion would be to:
- use INKSCAPE_COMMAND if available (should definitely work for Inkscape >= 1.3.x, even in AppImage)
- else determine INKSCAPEBIN from PATH as we did before (maybe needed for very very old Inkscape versions)
- do not implement anything special for AppImage (because they set INKSCAPE_COMMAND).

So basically replace all code that tries to determine the inkscape path with:
```
INKSCAPEBIN = os.environ.get("INKSCAPE_COMMAND") or which("inkscape", [INKSCAPEDIR])
```

If that doesn't work we can still add more complexity.

mgmax:
Hi  @tchilov, I see lots of new changes by you in this PR that seem helpful but are not related to the topic of this pull request. Feel free to create separate pull requests for these changes. Let's work together to get these into the mainline of VisiCut, so that all labs can profit from the development, and that your lab can profit from updates in mainline VisiCut. I'm happy to assist, in real life or here in the digital world.

tchilov:
@mgmax I didn't expect pushing to our master affecting this PR. Had I wanted to bring features upstream, I would have created feature branches and PRs.

mgmax:
I am temporarily closing this PR because it contains many commits from your "master" branch that have nothing to do with the PR itself. Every time I get new notifications. Feel free to reopen the PR against a separate branch. Would be cool to get the actual bugfix into VisiCut.

PR #722, opened 2024-04-08
Fix release version detection

TheAssassin:
Should fix build issues like in #721.

### DIFF
--- a/.github/workflows/github-build.yml
+++ b/.github/workflows/github-build.yml
@@ -31,6 +31,11 @@ jobs:
           submodules: "recursive"
           fetch-depth: 0
 
+      # continuous tag confuses git describe --tags, happens here specifically due to us fetching the entirety of the repository
+      - name: Delete continuous tag
+        run: |
+          git tag -d continuous || true
+
       - name: Build distribution
         run: |
           # we use a separate directory called output to "collect" all the build artifacts


.github/workflows/github-build.yml
```diff
@@ -31,6 +31,11 @@ jobs:
           submodules: "recursive"
           fetch-depth: 0
 
+      # continuous tag confuses git describe --tags, happens here specifically due to us fetching the entirety of the repository
+      - name: Delete continuous tag
+        run: |
+          git tag | grep '^continuous$' && git tag -d continuous
```
jnweiger, line 37:
```suggestion
          git tag -d continuous 2> /dev/null
```

Save some cpu cylces?

TheAssassin: It'll fail if the tag doesn't exist with return code 1, causing the step to fail. You'd have to `|| true` then. It won't really speed it up either, I guess. Also, allowing the output to end up in the log is handy for debugging.

TheAssassin: You could do the following:

```suggestion
          git tag -d continuous || true
```

mgmax:
Thanks for the improvement. I'm not sure if this really solves the root cause. We will probably have the same issue if someone clones and builds locally. Looks like GitHub releases is only intended for building "release" versions and not every single commit?

TheAssassin:
People are indeed affected locally as well. This was specifically meant to "fix" the problem on the CI for now. I know that @jnweiger has had this problem locally once already.

PR #727, opened 2024-04-27
Fix #720: Bounding box calculation for paths with transform

### LINKED ISSUE #720: Visicut shows extra padding around object
In inkscape I have drawn a numeral 3 (using hershey 1 line font)
The numeral is ca 4cm high, the viewbox fits very tightly.
For the screenshot, I pressed CTRL-A (select all) to demonstrate the effective bounding box: 

grafik

When this SVG is sent to visicut (either via extension or save and load) it is shown with a huge padding.
This makes it difficult to set the starting point of the laser accurately.

grafik

Expected behaviour:
* bunding box of the object in visicut is the exact bounding box. 
* viscut should not get confused by nested groups that stack several scalings and translations.

[bb33.svg.zip](https://github.com/t-oster/VisiCut/files/14849293/bb33.svg.zip)
bb33

mgmax:
For `<path transform="..." style="stroke-width:...">`, VisiCut ignored the `transform` when calculating the effective stroke width that is used to determine the bounding box.

As a result, some files showed way too large bounding boxes and could not be moved fully to the top left.

### DIFF
--- a/src/main/java/de/thomas_oster/visicut/model/graphicelements/GraphicObject.java
+++ b/src/main/java/de/thomas_oster/visicut/model/graphicelements/GraphicObject.java
@@ -28,6 +28,19 @@ import java.util.List;
  */
 public interface GraphicObject
 {
+  /**
+   * Get bounding box.
+   *
+   * The stroke width of lines is included in the bounding box (at least for SVG;
+   * the implementation status for other formats is unclear.)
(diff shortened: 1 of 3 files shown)


TheAssassin (commented):
Looks good. Have not run it yet.

mgmax:
Merging, as there was no opposite feedback for a longer time

